PR #16, opened 2020-05-29
PoC using redirect

### LINKED ISSUE #15: Implement a PoC version of the service
Write a minor PoC that allow perform the very basics operations that are expected for the smart-proxy.

# Acceptance criteria
- Have a service that runs locally and is able to respond same requests as aggregator and group service
- It can mock the data or redirects to the actual services in order to fetch real data

joselsegura:
# Description

Just a PoC for the smart-proxy service using HTTP redirects.

This solution doesn't cache anything, so the final smart-proxy will look very different from this implementation.

Fixes #15 

## Type of change

## Testing steps
Regular CI, no new tests at the moment

### DIFF
--- a/server/endpoints_test.go
+++ b/server/endpoints_test.go
@@ -28,5 +28,5 @@ func TestMakeURLToEndpointWithValidValue(t *testing.T) {
 
 	retval := server.MakeURLToEndpoint(apiPrefix, endpoint)
 
-	assert.Equal(t, retval, "api/v1/some_valid_endpoint")
+	assert.Equal(t, "api/v1/some_valid_endpoint", retval)
 }
--- a/server/endpoints_test.go
+++ b/server/endpoints_test.go
@@ -1,0 +1,32 @@
+// Copyright 2020 Red Hat, Inc
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package server_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/RedHatInsights/insights-results-smart-proxy/server"
+)
+
+func TestMakeURLToEndpointWithValidValue(t *testing.T) {
+	apiPrefix := "api/v1/"
+	endpoint := "some_valid_endpoint"
+
+	retval := server.MakeURLToEndpoint(apiPrefix, endpoint)
+
+	assert.Equal(t, retval, "api/v1/some_valid_endpoint")
+}
--- a/server/server.go
+++ b/server/server.go
@@ -233,9 +233,7 @@ func (server HTTPServer) proxyTo(baseURL string) func(http.ResponseWriter, *http
 			handleServerError(writer, err)
 		}
 		// Maybe this code should be on responses.SendRaw or something like that
-		writer.Header().Set("content-type", "application/json; charset=utf-8")
-		writer.WriteHeader(http.StatusOK)
-		_, err = writer.Write(content)
+		err = responses.Send(http.StatusOK, writer, content)
 		if err != nil {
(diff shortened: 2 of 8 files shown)


server/endpoints_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/RedHatInsights/insights-results-smart-proxy/server"
+)
+
+func TestMakeURLToEndpointWithValidValue(t *testing.T) {
+	apiPrefix := "api/v1/"
+	endpoint := "some_valid_endpoint"
+
+	retval := server.MakeURLToEndpoint(apiPrefix, endpoint)
+
+	assert.Equal(t, retval, "api/v1/some_valid_endpoint")
```
Sergey1011010, line 31: actually, expected is the first(doesn't really matter, just a nitpick)

server/server.go
```diff
@@ -171,3 +176,55 @@ func (server *HTTPServer) Start() error {
 func (server *HTTPServer) Stop(ctx context.Context) error {
 	return server.Serv.Shutdown(ctx)
 }
+
+// genericHandle
+func (server HTTPServer) genericAggregatorRedirect(writer http.ResponseWriter, request *http.Request) {
+	endpointComps := strings.Split(request.RequestURI, server.Config.APIPrefix)
```
Sergey1011010, line 182: maybe you meant strings.TrimPrefix?

server/server.go
```diff
@@ -171,3 +176,55 @@ func (server *HTTPServer) Start() error {
...
+	}
+
+	http.Redirect(writer, request, endpointURL.String(), 302)
+}
+
+// genericHandle
+func (server HTTPServer) genericContentServiceRedirect(writer http.ResponseWriter, request *http.Request) {
+	endpointComps := strings.Split(request.RequestURI, server.Config.APIPrefix)
+	endpoint := strings.Join(endpointComps[1:], "/")
+	endpointURL, err := url.Parse(server.ServicesConfig.ContentBaseEndpoint + endpoint)
```
Bee-lee, line 210: could this be moved into a function ?

server/server.go
natiiix: The result in case where the prefix is missing will be slightly different, but I agree that `TrimPrefix` is a safer choice here.

server/endpoints_test.go
joselsegura: Nice catch!

server/server.go
joselsegura: I'm new in strings handling in Go, so probably yes. TYVM

server/server.go
joselsegura: Working in it.
TYVM

server/server.go
```diff
@@ -171,3 +176,55 @@ func (server *HTTPServer) Start() error {
...
+		}
+
+		handleServerError(writer, err)
+	}
+
+	http.Redirect(writer, request, endpointURL.String(), 302)
+}
+
+// genericHandle
+func (server HTTPServer) genericContentServiceRedirect(writer http.ResponseWriter, request *http.Request) {
```
Sergey1011010, line 207: btw, genericAggregatorRedirect, and genericContentServiceRedirect are almost the same function, so maybe it could be something like

```golang
func (server HTTPServer) redirectTo(baseURL string) func(http.ResponseWriter, *http.Request)
```

tisnik:
> Looks good overall, I just don't like listing endpoints in each service, probably we could use prefixes, like
> 
> `/api/v1/aggregator/*` -> redirect to aggregator service
> `/api/v1/content/*` -> redirect to content service

the whole redirect things are basically throwaway code anyway

server/server.go
joselsegura: Done

joselsegura:
I added a new function able to perform the request to the aggregator/content-service and then response back to the client

server/server.go
```diff
@@ -171,3 +177,77 @@ func (server *HTTPServer) Start() error {
...
+		content, err := ioutil.ReadAll(response.Body)
+
+		if err != nil {
+			log.Error().Err(err).Msgf("Error while retrieving content from request to %s", endpointURL.String())
+			handleServerError(writer, err)
+		}
+		// Maybe this code should be on responses.SendRaw or something like that
+		writer.Header().Set("content-type", "application/json; charset=utf-8")
+		writer.WriteHeader(http.StatusOK)
+		writer.Write(content)
```
joselsegura, line 238: If PR [#56 of Operator Utils](https://github.com/RedHatInsights/insights-operator-utils/pull/56) is accepted and merged, this code should be refactored to a simply `responses.Send(http.StatusOK, writer, content)`

PR #26, opened 2020-06-05
Pull groups configuration from content-service

### LINKED ISSUE #18: Implement pull from insights-content-service and load groups configuration into memory
TL;DR
As explained in issue #7, the same idea should be implemented for groups configuration

The insights-content-service is already serving the groups configuration. The smart proxy should retrieve that configuration and update it periodically and store it in data structures in memory to being able to respond to group requests quickly

### Acceptance criteria

- Smart proxy is able to respond to `/groups` endpoint
- Smart proxy retrieves groups configuration from the `content-service` at start up
- Smart proxy retrieves periodically the groups from `content-service`, acting as cache when the service is not available

### Unknowns
- Which transport to use to avoid marshaling and unmarshaling same structures in both services

joselsegura:
# Description
This PR includes pulling the groups configuration from the content service and serving that configuration as an smart proxy endpoint.

It also includes a mechanism to periodically update the group configuration.

Fixes #18 

## Type of change

## Testing steps
Regular CI. Increased UT threshold from 0 to 23


config.toml
```diff
@@ -11,3 +11,4 @@ enable_cors = false
 [services]
 aggregator = "http://localhost:8080/api/v1/"
 content = "http://localhost:8082/api/v1/"
+groups_poll_time = 60
```
quarckster, line 14: Why do we need to poll?

joselsegura: We want to update the group configuration periodically to cache it. If we don't poll, we can miss groups configuration changes on the content service

joselsegura:
@tisnik good point, I didn't think about it. Do you mean to avoid collisions between reading the configuration and updating it from the content service?

Sergey1011010:
@tisnik but we have only 1 writing goroutine, don't we?

smart_proxy.go
```diff
@@ -100,6 +106,36 @@ func startServer() int {
 	return ExitStatusOK
 }
 
+// updateGroupInfo
```
tisnik, line 109: just to clarify (repeating what you wrote in IRC :):
please update comment a bit:
```
the updateGroupInfo **function** is run in a goroutine. It is always, in loop, waiting for one of 2 events: a ticket or a channel.
- if the ticket comes first, it tries to update the groups configuration from the content-service and, if success, it updates the current configuration
- if the channel comes first, then it send the latest retrieved configuration
```

joselsegura: I just updated the "doc strings" for `updateGroupInfo` and `getGroups`, where the channel is used

Sergey1011010:
I don't really understand why we need to put the same data to the channel in a loop. How is it better than just using a pointer?

server/server.go
```diff
@@ -241,6 +245,28 @@ func (server HTTPServer) proxyTo(baseURL string) func(http.ResponseWriter, *http
 	}
 }
 
+// getGroups retrives the groups configuration from a channel to get the latest valid one and send the response back to the client
+func (server *HTTPServer) getGroups(writer http.ResponseWriter, request *http.Request) {
+	select {
```
Sergey1011010, line 250: What's the purpose of this `select`? It's reading only from one channel, isn't it?

joselsegura: You are right, I don't remember why I put it, but it is not needed at all. TYVM!

joselsegura:
@Sergey1011010 the channel is used to pass the groups configuration from the "writer" goroutine (`updateGroupInfo`) to the "reader" one (`getGroup` HTTP request handler). Every time the reader needs a value, it can receive it from the channel, and using the channel makes the writer unable to write a new configuration (due a "tick") and send the configuration to the channel, so it acts as a synchronization mechanism

Sergey1011010:
@joselsegura, yes, but why do you need such a complicated mechanism when you can just replace an array at once in writing goroutine and you won't have any concurrency problems.

joselsegura:
After talking with @Sergey1011010, he convinced me that there is no possibility of having problems using a simple array mechanism, due to there will only be a goroutine writing the array, and given that the slice update operation is atomic, it cannot be possible to have a "readers/writer" problem here.

So, what do you think to returning back to the first proposal instead of using a channel?

tisnik:
> After talking with @Sergey1011010, he convinced me that there is no possibility of having problems using a simple array mechanism, due to there will only be a goroutine writing the array, and given that the slice update operation is atomic, it cannot be possible to have a "readers/writer" problem here.
> 
> So, what do you think to returning back to the first proposal instead of using a channel?

it's up to you (don't know about the discussion), but the official rule is simple:
- if multiple goroutines access a variable concurrently, and at least one of the accesses is a write, then synchronization is required.

And also according to spec:
- structured variables of array, slice, and struct types have elements and fields that may be addressed individually. Each such element acts like a variable.

Re atomicity: it's just slightly related to concurrent R/W operations.

joselsegura:
> it's up to you (don't know about the discussion), but the official rule is simple:
> 
>     * if multiple goroutines access a variable concurrently, and at least one of the accesses is a write, then synchronization is required.
> 
> 
> And also according to spec:
> 
>     * structured variables of array, slice, and struct types have elements and fields that may be addressed individually. Each such element acts like a variable.
> 
> 
> Re atomicity: it's just slightly related to concurrent R/W operations.

Apparently we are exactly matching the rule statement: we have several goroutines and one (and only one of them) is writing the values.

So, I will keep the current version until you find anything more to comment.

P.S. Plase, resolve the requests for change if you are ok with the updates

Sergey1011010:
@tisnik but slice stores a pointer to an array and updating this pointer(setting new array to a slice) is atomic, isn't it?

tisnik:
> @tisnik but slice stores a pointer to an array and updating this pointer(setting new array to a slice) is atomic, isn't it?

Indeed it might be on most architectures as max.capacity is equal to max `int` and `int` is "naturally" selected (32bit or 64bit). And slice stores also length, and capacity. We might assume it might be changed in atomic operation as well. But pointer+length+capacity (which is usually three words) are much more tricky.

joselsegura:
Wow, this scaled quickly :-D Golang internals! Poor Python guy

server/server.go
joselsegura: Are you ok with the comments included in the last version?

smart_proxy.go
joselsegura: Are you ok with the new comments in the last version?

server/server.go
joselsegura: Fixed in the last commit

smart_proxy.go
tisnik: yes, TY

PR #33, opened 2020-06-11
Endpoint for getting static content for a single rule

joselsegura:
# Description

Adding an endpoint into smart proxy that uses cached static content for rules to send back a JSON with all the static content for a given rule

Fixes [ccx-data-pipeline#261](https://github.com/RedHatInsights/ccx-data-pipeline/issues/261)

## Type of change

## Testing steps
Regular CI.

The new endpoint is in ${API_PREFIX}/content/{rule_id}, in order to check the returned value

### DIFF
--- a/server/endpoints.go
+++ b/server/endpoints.go
@@ -58,6 +58,8 @@ const (
 	EnableRuleForClusterEndpoint = "clusters/{cluster}/rules/{rule_id}/enable"
 	// MetricsEndpoint returns prometheus metrics
 	MetricsEndpoint = "metrics"
+	// RuleContent returns static content for {rule_id}
+	RuleContent = "content/{rule_id}"
 )
 
 func (server *HTTPServer) addDebugEndpointsToRouter(router *mux.Router) {
@@ -98,6 +100,7 @@ func (server *HTTPServer) addEndpointsToRouter(router *mux.Router) {
 	router.HandleFunc(apiPrefix+EnableRuleForClusterEndpoint, server.proxyTo(aggregatorEndpoint)).Methods(http.MethodPut, http.MethodOptions)
 	router.HandleFunc(apiPrefix+RuleGroupsEndpoint, server.getGroups).Methods(http.MethodGet, http.MethodOptions)
 	router.HandleFunc(apiPrefix+RuleErrorKeyEndpoint, server.proxyTo(aggregatorEndpoint)).Methods(http.MethodGet)
+	router.HandleFunc(apiPrefix+RuleContent, server.getContentForRule).Methods(http.MethodGet)
 
 	// Prometheus metrics
 	router.Handle(apiPrefix+MetricsEndpoint, promhttp.Handler()).Methods(http.MethodGet)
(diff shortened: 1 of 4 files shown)


server/handlers.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// getContentForRule retrieves the static content for the given ruleID
+func (server HTTPServer) getContentForRule(writer http.ResponseWriter, request *http.Request) {
+	contentConfig := <-server.ContentChannel
+
+	if contentConfig.Rules == nil {
+		err := errors.New("No rules content")
+		log.Error().Err(err).Msg("Rules static content cannot be retrieved from content service. Check logs")
+		handleServerError(writer, err)
```
tisnik, line 52: probably `return` is missing there?

server/handlers.go
```diff
@@ -0,0 +1,82 @@
...
+	ruleID, err := readRuleID(writer, request)
+	if err != nil {
+		// already handled in readRuleID
+		return
+	}
+
+	stringfiedRuleID := string(ruleID)
+
+	for _, ruleContent := range contentConfig.Rules {
+		if stringfiedRuleID != ruleContent.Plugin.PythonModule {
```
tisnik, line 64: probably it might be good to comment what does this condition means? :)

server/handlers.go
```diff
@@ -0,0 +1,82 @@
...
+			continue
+		}
+
+		err = responses.SendOK(writer, responses.BuildOkResponseWithData("content", ruleContent))
+		if err != nil {
+			log.Error().Err(err)
+			handleServerError(writer, err)
+		}
+		return
```
tisnik, line 73: is not the `return` misplaced?

tisnik: prob. not because of `continue`

server/handlers.go
joselsegura: You are right

server/handlers.go
joselsegura: I'll do

server/handlers.go
joselsegura: I preferred to use a continue and avoid one indentation level (placing the `SendOK` and error capture `if` inside an `else`). Probably my Python soul trying to avoid levels and levels of indentations :-D

Obviously the code is not clear enough, so I will change it to something more conventional

server/endpoints.go
```diff
@@ -58,6 +58,8 @@ const (
 	EnableRuleForClusterEndpoint = "clusters/{cluster}/rules/{rule_id}/enable"
 	// MetricsEndpoint returns prometheus metrics
 	MetricsEndpoint = "metrics"
+	// RuleContent returns static content for {rule_id}
+	RuleContent = "content/{rule_id}"
```
Sergey1011010, line 62: https://platform-docs.cloud.paas.psi.redhat.com/rest/urls.html#url-path-structure

maybe `rules/rule_id` or `rules/rule_id/content` would be slightly better with guidelines. Just a nitpick :)

joselsegura: Thank you for pointing this, I wasn't aware about that guidelines, so I will change it

PR #59, opened 2020-06-24
Don't install utils if already installed

### LINKED ISSUE #50: When `gosec` tool is already installed, skip the installation part
When `gosec` tool is already installed, skip the installation part

tisnik:
# Description

Don't install utils if already installed
(the same change that was incorporated into aggregator)

Fixes #50 #51 #52 #53 #54 #55 #56

## Type of change

## Testing steps

Can be checked by running all updated scripts - it is much much faster now.

### DIFF
--- a/abcgo.sh
+++ b/abcgo.sh
@@ -26,7 +26,11 @@ if [[ $* == *verbose* ]]; then
     VERBOSE=true
 fi
 
-GO111MODULE=off go get -u github.com/droptheplot/abcgo
+if ! [ -x "$(command -v abcgo)" ]
+then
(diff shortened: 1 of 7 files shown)


abcgo.sh
```diff
@@ -26,7 +26,11 @@ if [[ $* == *verbose* ]]; then
     VERBOSE=true
 fi
 
-GO111MODULE=off go get -u github.com/droptheplot/abcgo
+if ! [ -x "$(command -v abcgo)" ]
```
Sergey1011010, line 29: Then `-u` option doesn't make sense here

PR #73, opened 2020-06-30
Fix expected response on groups endpoint specification and some tooling

### LINKED ISSUE #72: /groups endpoint has incorrect status code in openapi spec
`/groups` endpoint has response 302 but in fact it returns 200 and it should fully be copied from https://github.com/RedHatInsights/insights-content-service/blob/master/openapi.json#L14
```json
    "/groups": {
      "get": {
        "summary": "Get all rule groups and their relevant information",
        "description": "This simply redirects to an endpoint of the same name of a service called insights-operator-service",
        "parameters": [],
        "operationId": "getRuleGroups",
        "responses": {
          "302": {
            "description": "Found redirect: response containing all rule groups",
            "content": {
              "text/plain": {}
            }
          },
```

joselsegura:
# Description

OpenAPI spec was outdated and refers to a redirection that is not done. Fixing and adding some tooling to manually check the OpenAPI specification

Fixes #72 

## Type of change

## Testing steps
Regular CI

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -70,6 +70,9 @@
         "summary": "Get all rule groups and their relevant information",
         "description": "This simply redirects to an endpoint of the same name of a service called insights-operator-service",
         "parameters": [],
+        "tags": [
+          "prod"
+        ],
         "operationId": "getRuleGroups",
         "responses": {
           "200": {
--- a/openapi.json
+++ b/openapi.json
@@ -72,10 +72,40 @@
         "parameters": [],
         "operationId": "getRuleGroups",
         "responses": {
-          "302": {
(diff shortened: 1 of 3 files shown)


openapi.json
```diff
@@ -72,10 +72,40 @@
         "parameters": [],
         "operationId": "getRuleGroups",
         "responses": {
-          "302": {
-            "description": "Found redirect: response containing all rule groups",
+          "200": {
```
quarckster, line 75: Please add `tags` stanza:
```json
"tags": ["prod"]
```

joselsegura: Done

PR #75, opened 2020-06-30
Taking into account some of the possible errors parsing an URL

### LINKED ISSUE #74: PUT /clusters/00000000000000000000000000000000000%1F/rules/0/[like,dislike,reset_vote] returns 500
Smart-proxy log excerpt:

```
{"level":"info","time":"2020-06-30T06:09:12Z","message":"Request received - URI: /api/insights-results-aggregator/v1/clusters/00000000000000000000000000000000000%1F/rules/0/like, Method: PUT"}
{"level":"error","error":"parse /api/insights-results-aggregator/v1/clusters/00000000000000000000000000000000000\u001f/rules/0/users/6089719/like: net/url: invalid control character in URL","time":"2020-06-30T06:09:12Z","message":"handleServerError()"}
```

joselsegura:
# Description

When an invalid character is received in a request, the endpoints that are redirected first try to parse the new URL before doing a request to the aggregator or content-service. when parsing, the method could fail and results in an "default case scenario" error, that results into a 500 status code. In this case, we should better response with a Bad Request response.

Fixes #74

## Type of change

## Testing steps
Regular CI

### DIFF
--- a/server/endpoints_test.go
+++ b/server/endpoints_test.go
@@ -81,3 +81,17 @@ func TestHTTPServer_ProxyTo_VoteEndpointsExtractUserID(t *testing.T) {
 }
 
 // TODO: test that proxying is done correctly including request / response modifiers for all endpoints
+
+func TestHTTPServer_ProxyTo_VoteEndpointBadCharacter(t *testing.T) {
+	badClusterName := "00000000000000000000000000000000000%1F"
+	helpers.AssertAPIRequest(t, &helpers.DefaultServerConfig, &helpers.DefaultServicesConfig, nil, &helpers.APIRequest{
+		Method:       http.MethodPut,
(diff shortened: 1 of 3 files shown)


server/endpoints_test.go
```diff
@@ -81,3 +81,17 @@ func TestHTTPServer_ProxyTo_VoteEndpointsExtractUserID(t *testing.T) {
 }
 
 // TODO: test that proxying is done correctly including request / response modifiers for all endpoints
+
+func TestHTTPServer_ProxyTo_VoteEndpointBadCharacter(t *testing.T) {
+	badClusterName := "00000000000000000000000000000000000%1F"
```
quarckster, line 86: I wish we have the same fuzzer in unit tests which I use in integration tests :)

PR #80, opened 2020-07-02
Modified the status code

### LINKED ISSUE #70: User from one organization can access to a list of clusters of another organization
After disabling authentication in insights-results-aggregator it's possible to access to data of an organization the user doesn't belong to. If I enable authentication back GET `/organizations/{orgId}/clusters` will return 403 with error `Missing authentication token`.

Steps to reproduce:
1. Upload an archive using the credentials of one organization.
2. Perform GET `/organizations/{org id from step 1}/clusters` with the credentials of another organization

Actual result:
A list of clusters is received and the status code is 200.

Expected results:
Status code 403 or 404

joselsegura:
# Description

Change the status code returned from 403 to 404

Fixes #70 

## Type of change

## Testing steps
Regular CI

### DIFF
--- a/server/router_utils.go
+++ b/server/router_utils.go
@@ -131,6 +131,8 @@ func readOrganizationID(writer http.ResponseWriter, request *http.Request, auth 
 	return types.OrgID(organizationID), err
 }
 
+// checkPermissions checks if the provided organization ID matches with the one of the authenticated user
+// when the authentication is enabled
 func checkPermissions(writer http.ResponseWriter, request *http.Request, orgID types.OrgID, auth bool) error {
 	identityContext := request.Context().Value(types.ContextKeyUser)
 	if identityContext != nil && auth {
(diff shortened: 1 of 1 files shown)


server/router_utils.go
```diff
@@ -131,14 +131,16 @@ func readOrganizationID(writer http.ResponseWriter, request *http.Request, auth
...
+// when the authentication is enabled
 func checkPermissions(writer http.ResponseWriter, request *http.Request, orgID types.OrgID, auth bool) error {
 	identityContext := request.Context().Value(types.ContextKeyUser)
 	if identityContext != nil && auth {
 		identity := identityContext.(types.Identity)
 		if identity.Internal.OrgID != orgID {
 			const message = "You have no permissions to get or change info about this organization"
 			log.Error().Msg(message)
-			handleServerError(writer, &AuthenticationError{errString: message})
+			handleServerError(writer, &types.ItemNotFoundError{ItemID: orgID})
```
Sergey1011010, line 143: Why do you return item not found when it's authentication error?

quarckster: For sake of consistency. In this issue https://github.com/RedHatInsights/insights-results-aggregator/issues/907 I raised that question and we agreed to return 404. This issue https://github.com/RedHatInsights/insights-results-smart-proxy/issues/70 is about the same.

Sergey1011010: @quarckster I commented there that we can return 403 and it doesn't make sense to return 404 when it's obviously an authentication error. Well, anyway, it's not a big deal, some apis return 404 instead of 403 as well.

quarckster: > some apis return 404 instead of 403 as well.

Yes, and we decided just to follow that pattern.

PR #96, opened 2020-07-14
Update stable

quarckster:


quarckster:
e2e test failed due to outdated dependency in CI, I ran it locally and it passed.

PR #137, opened 2020-07-24
New endpoint returning all rule IDs

joselsegura:
# Description

Adding a /rules endpoint to retrieve the list of rule IDs. It also checks the organization ID to retrieve only external rules or include the internal ones too

## Type of change

## Testing steps

regular CI


server/endpoints.go
```diff
@@ -34,6 +34,8 @@ const (
 	RuleGroupsEndpoint = "groups"
 	// RuleContent returns static content for {rule_id}
 	RuleContent = "rules/{rule_id}/content"
+	// RuleIDs returns a list of rule IDs
+	RuleIDs = "rules"
```
Sergey1011010, line 38: Maybe rule_ids?

server/handlers.go
```diff
@@ -61,9 +62,9 @@ func (server HTTPServer) getContentForRule(writer http.ResponseWriter, request *
 
 	// check for internal rule permissions
 	if internal := content.IsRuleInternal(ruleID); internal == true {
-		ok := server.checkInternalRulePermissions(writer, request)
+		ok, err := server.checkInternalRulePermissions(request)
 		if ok != true {
```
Sergey1011010, line 66: you meant `err != nil`?

server/handlers.go
```diff
@@ -87,3 +88,25 @@ func (server HTTPServer) getClustersForOrg(writer http.ResponseWriter, request *
 	server.proxyTo(server.ServicesConfig.AggregatorBaseEndpoint, nil)(writer, request)
 	return
 }
+
+// getRuleIDs returns a list of the names of the rules
+func (server HTTPServer) getRuleIDs(writer http.ResponseWriter, request *http.Request) {
+	allRuleIDs := content.GetRuleIDs()
+	var ruleIDs []string
+
+	if hasPermissions, _ := server.checkInternalRulePermissions(request); !hasPermissions {
```
Sergey1011010, line 97: Is the last param error? Why do you ignore it if so?

server/endpoints.go
tisnik: +1

joselsegura: ACK, I will change it

server/handlers.go
joselsegura: I was checking the `checkInternalRulePermission` and I realized that returning the `error` will be enough, because the `bool` value is always `true` when `error` is `nil`, and `false` if `error != nil`, so probably I will refactor the method and the callers. Thank you very much!

PR #141, opened 2020-07-27
Use specific type for exit code

tisnik:
# Description

Use specific type for exit code

## Type of change

## Testing steps

N/A

### DIFF
--- a/smart_proxy.go
+++ b/smart_proxy.go
@@ -37,6 +37,10 @@ import (
 	proxy_content "github.com/RedHatInsights/insights-results-smart-proxy/content"
 )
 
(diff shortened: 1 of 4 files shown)


smart_proxy.go
```diff
@@ -37,6 +37,10 @@ import (
 	proxy_content "github.com/RedHatInsights/insights-results-smart-proxy/content"
 )
 
+// ExitCode represents numeric value returned to parent process when the
+// current process finishes
+type ExitCode int
+
 const (
 	// ExitStatusOK means that the service have finished with success
 	ExitStatusOK = iota
```
Sergey1011010, line 46: I think we could use ExitCode as a type here

PR #155, opened 2020-07-31
Fix #151

### LINKED ISSUE #151: Wrong unit test causes CI to fail
```
--- FAIL: TestRuleNamesResponse (0.00s)
    strings.go:50: 
        	Error Trace:	strings.go:50
        	            				http.go:124
        	            				http.go:89
        	            				http.go:110
        	            				server_test.go:372
        	            				timeout.go:33
        	            				asm_amd64.s:1357
        	Error:      	Not equal: 
        	            	expected: map[string]interface {}{"rules":[]interface {}{"foo.rules.internal.bar", "ccx_rules_ocp.external.rules.node_installer_degraded"}, "status":"ok"}
        	            	actual  : map[string]interface {}{"rules":[]interface {}{"ccx_rules_ocp.external.rules.node_installer_degraded", "foo.rules.internal.bar"}, "status":"ok"}
      	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -2,4 +2,4 @@
        	            	  (string) (len=5) "rules": ([]interface {}) (len=2) {
        	            	-  (string) (len=22) "foo.rules.internal.bar",
        	            	-  (string) (len=52) "ccx_rules_ocp.external.rules.node_installer_degraded"
        	            	+  (string) (len=52) "ccx_rules_ocp.external.rules.node_installer_degraded",
        	            	+  (string) (len=22) "foo.rules.internal.bar"
        	            	  },
        	Test:       	TestRuleNamesResponse
        	Messages:   	{"rules": ["foo.rules.internal.bar", "ccx_rules_ocp.external.rules.node_installer_degraded"],"status": "ok"}
        	            	and
        	            	{"rules":["ccx_rules_ocp.external.rules.node_installer_degraded","foo.rules.internal.bar"],"status":"ok"}
        	            	should represent the same json
FAIL
```

joselsegura:
# Description

The problem is caused because the list returned by the `rule_ids` endpoint  is generated from a `map`, so its order is not granted.
This fix orders the output using `sort`, making the result predictable

Fixes #151 

## Type of change

## Testing steps

Regular CI

### DIFF
--- a/content/content.go
+++ b/content/content.go
@@ -18,7 +18,6 @@ package content
 
 import (
 	"fmt"
-	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -122,7 +121,6 @@ func (s *RulesWithContentStorage) GetRuleIDs() []string {
 		ruleIDs = append(ruleIDs, ruleContent.Plugin.PythonModule)
 	}
 
-	sort.Strings(ruleIDs)
 	return ruleIDs
 }
 
--- a/content/content.go
+++ b/content/content.go
@@ -18,6 +18,7 @@ package content
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -121,6 +122,7 @@ func (s *RulesWithContentStorage) GetRuleIDs() []string {
 		ruleIDs = append(ruleIDs, ruleContent.Plugin.PythonModule)
 	}
 
+	sort.Strings(ruleIDs)
 	return ruleIDs
 }
 
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -17,8 +17,10 @@ limitations under the License.
 package server_test
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
+	"sort"
 	"testing"
 	"time"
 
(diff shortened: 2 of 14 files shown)


content/content.go
```diff
@@ -121,6 +122,7 @@ func (s *RulesWithContentStorage) GetRuleIDs() []string {
 		ruleIDs = append(ruleIDs, ruleContent.Plugin.PythonModule)
 	}
 
+	sort.Strings(ruleIDs)
```
Sergey1011010, line 125: I think it's better to sort when comparing in tests

joselsegura: I think about it and you are right. I just modified it

server/server_test.go
```diff
@@ -390,8 +392,32 @@ func TestRuleNamesResponse(t *testing.T) {
...
+	}
+
+	if err := json.Unmarshal(got, &gotResp); err != nil {
+		err = fmt.Errorf(`"got" is not JSON. value = "%v", err = "%v"`, got, err)
+		helpers.FailOnError(t, err)
+	}
+
+	sort.Strings(expectedResp.Rules)
+	sort.Strings(gotResp.Rules)
+	assert.Equal(t, expectedResp, gotResp)
```
Sergey1011010, line 422: btw, there some function in assert package which does the same as these 3 lines

joselsegura: Mmm, interesting. I discovered a lot of new asserting functions on `assert` package, but I cannot find the one you are referring to.

The most likely for me is using 2 `assert.Subset` (`expectedResp` is a subset of `gotResp` and vice versa).

joselsegura: `assert.ElementsMatch` is the answer. Nothing like writting a comment asking for something to find it by yourself

joselsegura: Updated commit using `assert.ElementsMatch` instead of sorting by myself

PR #207, opened 2020-08-07
New overview endpoint

### LINKED ISSUE #132: REST API endpoint to be consumed by "Aggregated Insights information in OCM" page
# Task
Prepare REST API endpoint to be consumed by "Aggregated Insights information in OCM" page

## Background
The mock up for the cluster overview page is prepared.

### Input

OCM UI provides `org_id`

### Output
- number of clusters with >1 insights hit
- number of clusters with hits of criticality X,Y,Z
- number of clusters with hitting health check for category X,Y,Z.

## OCM UI design

https://marvelapp.com/prototype/852jaj9/screen/70947041

joselsegura:
# Description

New endpoint for getting an overview about the number of clusters that are being impacted by rules

**Need to refactor because the cyclomatic complexity (so, CI should fail)**

Fixes #132 

## Type of change

- REST API tests
- Documentation update

## Testing steps

N/A for the moment. OpenAPI spec manually checked using the OpenAPI tools validator


openapi.json
```diff
@@ -1380,6 +1380,61 @@
           "prod"
         ]
       }
+    },
+    "/overview": {
+      "get": {
+        "summary": "Get relevant data for an overview of hitting rules in the clusters for this irganziation",
```
tisnik, line 1386: _i_rginazation

openapi.json
```diff
@@ -1380,6 +1380,61 @@
           "prod"
         ]
       }
+    },
+    "/overview": {
```
Sergey1011010, line 1384: Maybe we could consider to make the name somehow more related to the rules

openapi.json
```diff
@@ -1380,6 +1380,61 @@
...
+      "get": {
+        "summary": "Get relevant data for an overview of hitting rules in the clusters for this organziation",
```
Sergey1011010, line 1386: I don't see any parameters. How do we know the cluster name?

openapi.json
Sergey1011010, line 1384: Maybe "org_overview" or something like this?

server/handlers.go
```diff
@@ -134,3 +134,81 @@ func (server HTTPServer) getRuleIDs(writer http.ResponseWriter, request *http.Re
...
+	if err != nil {
+		handleServerError(writer, err)
+		return
+	}
+
+	userID := authToken.AccountNumber
+	orgID := authToken.Internal.OrgID
+
+	clustersHits := 0
+	var hitsByTotalRisk map[int]int = make(map[int]int)
```
Sergey1011010, line 150: you actually don't need `var` and `map[int]int` here :)

server/handlers.go
```diff
@@ -134,3 +134,81 @@ func (server HTTPServer) getRuleIDs(writer http.ResponseWriter, request *http.Re
...
+			log.Info().Msgf("Aggregator doesn't have reports for cluster ID %s", clusterID)
+			continue
+		}
+
+		if aggregatorResponse.Meta.Count == 0 {
+			continue
+		}
+		clustersHits = clustersHits + 1
+
+		var cacheTotalRisk map[int]bool = make(map[int]bool)
```
Sergey1011010, line 171: Didn't really get the point of these variables

server/handlers.go
```diff
@@ -134,3 +134,81 @@ func (server HTTPServer) getRuleIDs(writer http.ResponseWriter, request *http.Re
...
+			ruleID := report.Module
+			errorKey := report.ErrorKey
+
+			ruleWithContent, err := content.GetRuleWithErrorKeyContent(ruleID, errorKey)
+			if err != nil {
+				handleServerError(writer, err)
+				return
+			}
+
+			if val, found := cacheTotalRisk[ruleWithContent.TotalRisk]; !found || val {
```
Sergey1011010, line 183: What's the point of this? It's always true, isn't it?

server/server.go
```diff
@@ -332,6 +332,33 @@ func copyHeader(srcHeaders http.Header, dstHeaders http.Header) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	type clustersResponse struct {
+		Status   string              `json:"status"`
+		Clusters []types.ClusterName `json:"clusters"`
+	}
+
+	var recvMsg clustersResponse
```
Sergey1011010, line 354: btw, you could omit type declaration here 
```
var recvMsg struct {
...
```

server/server.go
```diff
@@ -332,6 +332,33 @@ func copyHeader(srcHeaders http.Header, dstHeaders http.Header) {
...
+	if err = json.NewDecoder(response.Body).Decode(&recvMsg); err != nil {
```
Sergey1011010, line 355: Could be shortened like this
```
err := json.NewDecoder(response.Body).Decode(&recvMsg)
return recvMsg.Clusters, err
```

openapi.json
joselsegura: Good idea

server/handlers.go
joselsegura: Ack

server/handlers.go
joselsegura: As the overview intends to show the number of clusters affected by each level of "criticality" (calculated total risk), I need to ensure that I don't count for each total risk value the same cluster twice.

This map allows me to know if this total risk has been already counted for this cluster.

OTOH, as I need to refactor this function to fulfil our cyclomatic expectations, probably I will come with a different solution and maybe I can avoid this kind of cache variables

server/handlers.go
joselsegura: When you iterate for a given report in a given cluster, at least the first hit won't be already in this map. Both `cache*` vars will be re-created when the first `for` iterate to the next cluster

server/server.go
joselsegura: TYVM, I will fix it

server/server.go
joselsegura: Good point, tyvm

server/handlers.go
Sergey1011010: And what's the point? You don't use this cache anywhere and your `if` is hit every time. Or am I missing something?

joselsegura: The cache is used right here just to control when the `if` block is executed or not. Let me put an example: for a given cluster, you have a report with 3 rule hits. 1st and 3rd have the same "total risk" value, and 2nd a different one.

When you iterate over the rules in the report, first time you will check the `cacheTotalRisk`, you won't find the key and execute the `if` block, that sets the value in the `cacheTotalRisk` map and increase de real counter.

For the 2nd rule, the same will apply because the "total risk" is different.

For the 3rd rule, you will check the `cacheTotalRisk` and you will find a `true` value, so you don't enter inside the `if` block and avoid counting the same risk 2 times for this cluster

Issue #62, opened 2020-06-24
Use single rule endpoint from aggregator


bond95:
This task is depend on https://github.com/RedHatInsights/insights-results-aggregator/issues/896 . When that issue fill be done, smart proxy should use that created endpoint for single rule.

Acceptance criteria:
 - Use single rule endpoint instead of report endpoint

PR #211, opened 2020-08-11
Replaced loop in single rule endpoint with request

bond95:
# Description

Changed loop with new single rule endpoint on aggregator side.

Fixes https://github.com/RedHatInsights/insights-results-smart-proxy/issues/62

## Type of change

## Testing steps

`make test`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,42 +7,17 @@ require (
 	github.com/RedHatInsights/insights-content-service v0.0.0-20201006090430-1fa402db0eb8
 	github.com/RedHatInsights/insights-operator-utils v1.6.1
 	github.com/RedHatInsights/insights-results-aggregator v1.1.0
-	github.com/RedHatInsights/insights-results-aggregator-data v0.0.0-20200826060335-2d341f371b98
-	github.com/Shopify/sarama v1.27.0 // indirect
-	github.com/aws/aws-sdk-go v1.35.3 // indirect
+	github.com/RedHatInsights/insights-results-aggregator-data v0.0.0-20201009114602-20a158ff4960
 	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869
 	github.com/deckarep/golang-set v1.7.1
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
-	github.com/fsnotify/fsnotify v1.4.9 // indirect
-	github.com/getkin/kin-openapi v0.22.1 // indirect
-	github.com/golang/snappy v0.0.2 // indirect
 	github.com/google/uuid v1.1.2
 	github.com/gorilla/handlers v1.4.2
 	github.com/gorilla/mux v1.8.0
-	github.com/klauspost/compress v1.11.1 // indirect
 	github.com/lib/pq v1.8.0 // indirect
-	github.com/magiconair/properties v1.8.4 // indirect
-	github.com/mitchellh/mapstructure v1.3.3 // indirect
-	github.com/pelletier/go-toml v1.8.1 // indirect
-	github.com/pierrec/lz4 v2.5.2+incompatible // indirect
 	github.com/prometheus/client_golang v1.7.1
-	github.com/prometheus/common v0.14.0 // indirect
-	github.com/prometheus/procfs v0.2.0 // indirect
-	github.com/rcrowley/go-metrics v0.0.0-20200313005456-10cdbea86bc0 // indirect
 	github.com/rs/zerolog v1.20.0
-	github.com/spf13/afero v1.4.1 // indirect
-	github.com/spf13/cast v1.3.1 // indirect
-	github.com/spf13/jwalterweatherman v1.1.0 // indirect
-	github.com/spf13/pflag v1.0.5 // indirect
 	github.com/spf13/viper v1.7.1
 	github.com/stretchr/testify v1.6.1
-	golang.org/x/crypto v0.0.0-20201002170205-7f63de1d35b0 // indirect
-	golang.org/x/net v0.0.0-20201002202402-0a1ea396d57c // indirect
-	golang.org/x/sys v0.0.0-20201005172224-997123666555 // indirect
-	golang.org/x/text v0.3.3 // indirect
-	google.golang.org/protobuf v1.25.0 // indirect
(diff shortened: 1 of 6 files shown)


server/server.go
```diff
@@ -381,6 +381,56 @@ func (server HTTPServer) readAggregatorReportForClusterID(
...
+func (server HTTPServer) readAggregatorRuleForClusterID(
+	orgID types.OrgID, clusterID types.ClusterName, userID types.UserID, ruleID types.RuleID, errorKey types.ErrorKey, writer http.ResponseWriter,
+) (*types.RuleOnReport, bool) {
+	aggregatorURL := httputils.MakeURLToEndpoint(
+		server.ServicesConfig.AggregatorBaseEndpoint,
+		ira_server.RuleEndpoint,
+		orgID,
+		clusterID,
+		userID,
+		fmt.Sprintf("%v|%v", ruleID, errorKey),
```
Sergey1011010, line 396: Why to split using "|"? You can add a parameter to the endpoint

tisnik: IMHO it is based on the original design of Insights for RHEL. They choose that weird thing for some reason :1st_place_medal:

bond95: Yes, it is based on Insights style.

go.mod
```diff
@@ -35,3 +35,5 @@ require (
 	gopkg.in/h2non/gock.v1 v1.0.15
 	gopkg.in/ini.v1 v1.57.0 // indirect
 )
+
+replace github.com/RedHatInsights/insights-results-aggregator => ../insights-results-aggregator
```
Sergey1011010, line 39: Please remove

joselsegura (approved):

server/handlers_test.go
```diff
@@ -67,6 +68,49 @@ func TestHTTPServer_ReportEndpoint(t *testing.T) {
...
+
+		helpers.GockExpectAPIRequest(t, helpers.DefaultServicesConfig.ContentBaseEndpoint, &helpers.APIRequest{
+			Method:   http.MethodGet,
+			Endpoint: ics_server.AllContentEndpoint,
+		}, &helpers.APIResponse{
+			StatusCode: http.StatusOK,
+			Body:       helpers.MustGobSerialize(t, testdata.RuleContentDirectory3Rules),
+		})
+
+		go content.RunUpdateContentLoop(helpers.DefaultServicesConfig)
```
tisnik, line 98: how is this supposed to work?

server/router_utils.go
```diff
@@ -215,6 +215,47 @@ func readOrganizationIDs(writer http.ResponseWriter, request *http.Request) ([]t
...
+			paramName:  "rule_id",
+			paramValue: ruleIDWithErrorKey,
+			errString:  err.Error(),
+		})
+		return types.RuleID(0), types.ErrorKey(0), err
+	}
+
+	IDValidator := regexp.MustCompile(`^[a-zA-Z_0-9.]+$`)
+
+	isRuleIDValid := IDValidator.Match([]byte(splitedRuleID[0]))
```
tisnik, line 242: what about `MatchString`?

server/router_utils.go
```diff
@@ -215,6 +215,47 @@ func readOrganizationIDs(writer http.ResponseWriter, request *http.Request) ([]t
...
+	isErrorKeyValid := IDValidator.Match([]byte(splitedRuleID[1]))
+
+	if !isRuleIDValid || !isErrorKeyValid {
+		err = fmt.Errorf("invalid rule ID, each part of ID must contain only from latin characters, number, underscores or dots")
```
tisnik, line 246: please remove `from`

server/router_utils.go
```diff
@@ -215,6 +215,47 @@ func readOrganizationIDs(writer http.ResponseWriter, request *http.Request) ([]t
...
+		log.Error().Err(err)
+		handleServerError(writer, &RouterParsingError{
+			paramName:  "rule_id",
+			paramValue: ruleIDWithErrorKey,
+			errString:  err.Error(),
+		})
+		return types.RuleID(0), types.ErrorKey(0), err
+	}
+
+	IDValidator := regexp.MustCompile(`^[a-zA-Z_0-9.]+$`)
```
tisnik, line 240: is `0`, `.`, `-` accepted ID or error key?

server/handlers_test.go
bond95: Same as here https://github.com/RedHatInsights/insights-results-smart-proxy/pull/211/files#diff-e5b0e65ddaf27a82498a9a19be9839c2L54 . It's updating rules content (groups, templates and etc) by timer.

server/router_utils.go
bond95: Yes, for ID mostly used `a-z` and `.`, in error key are mostly used `A-Z` and `_`, but it could also be `a-z` and etc.

tisnik: I mean, with the proposed regexp, even single `.` will be accepted. Or `..` etc.

bond95: yes, and we use this pattern not only here, but in other places as well, to check ID. So I think it should be another refactor task

server/server.go
```diff
@@ -538,24 +588,68 @@ func (server HTTPServer) findRule(
...
+		Generic:      ruleWithContent.Generic,
+		Reason:       ruleWithContent.Reason,
+		Resolution:   ruleWithContent.Resolution,
+		TotalRisk:    ruleWithContent.TotalRisk,
+		RiskOfChange: ruleWithContent.RiskOfChange,
+		RuleID:       aggregatorResponse.Module,
+		TemplateData: aggregatorResponse.TemplateData,
+		Tags:         ruleWithContent.Tags,
+		UserVote:     aggregatorResponse.UserVote,
+		Disabled:     aggregatorResponse.Disabled,
```
Sergey1011010, line 649: looks like it lacks DisabledFeedback and DisabledAt attributes https://github.com/RedHatInsights/insights-results-smart-proxy/blob/master/server/server.go#L522 Even better would be to make some converting function probably in proxy_types package so that we have only one place to edit

Sergey1011010 (changes requested):
only one thing mentioned above

Sergey1011010 (approved):

PR #212, opened 2020-08-13
Fix issue #174

### LINKED ISSUE #174: Rules is always nil
image
https://github.com/RedHatInsights/insights-results-smart-proxy/blob/master/server/handlers.go#L87

joselsegura:
# Description

Minor bug regarding content endpoint

Fixes #174

## Type of change

## Testing steps

Regular CI


joselsegura:
@tisnik you are right as usual :-D

I just checked that the `getContent` handler has a 0% of coverage, so I will add some unit test to fix it

PR #303, opened 2020-10-05
ACM interface push model sequence diagram

### LINKED ISSUE #265: Create diagram with description how to push data to ACM using the "stream" approach instead of pulling
Create diagram with description how to push data to ACM using the "stream" approach instead of pulling

tisnik:
# Description

ACM interface push model sequence diagram

Fixes #265

## Type of change

- Documentation update

## Testing steps

N/A


joselsegura (approved):
My only question about the diagram is:
How does smart proxy know which clusters should be pushed to each ACM client?

tisnik:
> My only question about the diagram is:
> How does smart proxy know which clusters should be pushed to each ACM client?

The ACM knows timestamp of last result. It needs to specify that timestamp in the "start streaming" request. Then smart proxy simply starts to push all records from this timestamp (+- some delta, just not to miss anything).

joselsegura:
> > My only question about the diagram is:
> > How does smart proxy know which clusters should be pushed to each ACM client?
> 
> The ACM knows timestamp of last result. It needs to specify that timestamp in the "start streaming" request. Then smart proxy simply starts to push all records from this timestamp (+- some delta, just not to miss anything).

I just remembered about OrgID. I could imagine that is the dot I missed on this approach.

TYVM!

tisnik:
> > > My only question about the diagram is:
> > > How does smart proxy know which clusters should be pushed to each ACM client?
> > 
> > 
> > The ACM knows timestamp of last result. It needs to specify that timestamp in the "start streaming" request. Then smart proxy simply starts to push all records from this timestamp (+- some delta, just not to miss anything).
> 
> I just remembered about OrgID. I could imagine that is the dot I missed on this approach.
> 
> TYVM!

ACM has to info about org ID right now, so for demo+benchmarking purposes we decided not to use it (and not really needed for one ACM hub)

Bee-lee (approved):

PR #316, opened 2020-10-20
Added OSD tag support

bond95:
# Description

With this feature is able to return only rules that could be fixed on OSD clusters.

## Type of change

## Testing steps

`make test`

### DIFF
--- a/content/content.go
+++ b/content/content.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/RedHatInsights/insights-operator-utils/types"
+	local_types "github.com/RedHatInsights/insights-results-smart-proxy/types"
 	"github.com/rs/zerolog/log"
 
 	"github.com/RedHatInsights/insights-results-smart-proxy/services"
@@ -43,7 +44,7 @@ type ruleIDAndErrorKey struct {
 // It's thread safe
 type RulesWithContentStorage struct {
 	sync.RWMutex
-	rulesWithContent map[ruleIDAndErrorKey]*types.RuleWithContent
+	rulesWithContent map[ruleIDAndErrorKey]*local_types.RuleWithContent
 	rules            map[types.RuleID]*types.RuleContent
 }
 
@@ -57,7 +58,7 @@ func SetRuleContentDirectory(contentDir *types.RuleContentDirectory) {
 // GetRuleWithErrorKeyContent returns content for rule with error key
 func (s *RulesWithContentStorage) GetRuleWithErrorKeyContent(
 	ruleID types.RuleID, errorKey types.ErrorKey,
-) (*types.RuleWithContent, bool) {
+) (*local_types.RuleWithContent, bool) {
 	s.RLock()
 	defer s.RUnlock()
 
@@ -92,7 +93,7 @@ func (s *RulesWithContentStorage) GetAllContent() []types.RuleContent {
 
 // SetRuleWithContent sets content for rule with error key
 func (s *RulesWithContentStorage) SetRuleWithContent(
-	ruleID types.RuleID, errorKey types.ErrorKey, ruleWithContent *types.RuleWithContent,
+	ruleID types.RuleID, errorKey types.ErrorKey, ruleWithContent *local_types.RuleWithContent,
 ) {
 	s.Lock()
 	defer s.Unlock()
@@ -118,7 +119,7 @@ func (s *RulesWithContentStorage) ResetContent() {
 	s.Lock()
 	defer s.Unlock()
 
-	s.rulesWithContent = make(map[ruleIDAndErrorKey]*types.RuleWithContent)
+	s.rulesWithContent = make(map[ruleIDAndErrorKey]*local_types.RuleWithContent)
 	s.rules = make(map[types.RuleID]*types.RuleContent)
 }
 
@@ -137,7 +138,7 @@ func (s *RulesWithContentStorage) GetRuleIDs() []string {
 }
 
 var rulesWithContentStorage = RulesWithContentStorage{
-	rulesWithContent: map[ruleIDAndErrorKey]*types.RuleWithContent{},
+	rulesWithContent: map[ruleIDAndErrorKey]*local_types.RuleWithContent{},
 	rules:            map[types.RuleID]*types.RuleContent{},
 }
 
@@ -156,7 +157,7 @@ func WaitForContentDirectoryToBeReady() {
 // Caching is done under the hood, don't worry about it.
 func GetRuleWithErrorKeyContent(
 	ruleID types.RuleID, errorKey types.ErrorKey,
-) (*types.RuleWithContent, error) {
+) (*local_types.RuleWithContent, error) {
 	// to be sure the data is there
 	WaitForContentDirectoryToBeReady()
 
--- a/content/parsing.go
+++ b/content/parsing.go
@@ -18,7 +18,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/RedHatInsights/insights-operator-utils/collections"
 	"github.com/RedHatInsights/insights-operator-utils/types"
+	local_types "github.com/RedHatInsights/insights-results-smart-proxy/types"
 	"github.com/rs/zerolog/log"
 )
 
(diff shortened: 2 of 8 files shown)


content/content.go
```diff
@@ -43,7 +44,7 @@ type ruleIDAndErrorKey struct {
 // It's thread safe
 type RulesWithContentStorage struct {
 	sync.RWMutex
-	rulesWithContent map[ruleIDAndErrorKey]*types.RuleWithContent
+	rulesWithContent map[ruleIDAndErrorKey]*local_types.RuleWithContent
```
Sergey1011010, line 47: why did you change this?

server/server.go
```diff
@@ -458,7 +459,7 @@ func (server HTTPServer) readAggregatorRuleForClusterID(
 	return aggregatorResponse.Report, true
 }
 
-func (server HTTPServer) fetchRuleContent(rule types.RuleOnReport) (proxy_types.RuleWithContentResponse, bool) {
+func (server HTTPServer) fetchRuleContent(rule types.RuleOnReport, OSDEligible bool) (proxy_types.RuleWithContentResponse, bool, bool) {
```
Sergey1011010, line 462: I think it's better to document the return options since it's not very obvious what they are without reading the code

server/server.go
```diff
@@ -517,16 +522,32 @@ func (server HTTPServer) fetchAggregatorReport(
...
 	if !successful {
 		return
 	}
 
 	var rules []proxy_types.RuleWithContentResponse
+	hasNoErrors := true
 
 	for _, aggregatorRule := range aggregatorResponse.Report {
-		rule, successful := server.fetchRuleContent(aggregatorRule)
+		rule, successful, isError := server.fetchRuleContent(aggregatorRule, server.getOSDFlag(request))
```
Sergey1011010, line 548: what's the difference between successful and isError?

server/server.go
```diff
@@ -535,13 +556,16 @@ func (server HTTPServer) reportEndpoint(writer http.ResponseWriter, request *htt
 		rules = append(rules, rule)
 	}
 
-	if len(aggregatorResponse.Report) != 0 && len(rules) == 0 {
+	if len(aggregatorResponse.Report) != 0 && !hasNoErrors {
```
Sergey1011010, line 559: why?

server/server.go
```diff
@@ -591,14 +615,19 @@ func (server HTTPServer) singleRuleEndpoint(writer http.ResponseWriter, request
 		return
 	}
 
-	rule, successful = server.fetchRuleContent(*aggregatorResponse)
+	rule, successful, _ = server.fetchRuleContent(*aggregatorResponse, server.getOSDFlag(request))
 
 	if !successful {
+		err := responses.SendOK(writer, responses.BuildResponse("Rule was not found"))
```
Sergey1011010, line 621: How can it be 200 when the rule wasn't found?

content/content.go
bond95: Because I use local `RuleWithContent` which is used only in smart-proxy

server/server.go
bond95: Yes, sure

server/server.go
bond95: `successful` indicates if rule content was fetched (it could be filtered because of OSD flag). `isError` indicate if there appear some error, for example during parsing.

server/server.go
bond95: Because `len(rules)` could be empty because of filtering, but also because of `errors`, so this `if` should be triggered only in case of errors.

server/server.go
bond95: Oops, you're right

server/server.go
tisnik: dunno if it is not better to return error value with proper information about the failure

server/server.go
tisnik: I'd prefer to use `error` instead, WDYT?

server/server.go
```diff
@@ -458,7 +459,12 @@ func (server HTTPServer) readAggregatorRuleForClusterID(
 	return aggregatorResponse.Report, true
 }
 
-func (server HTTPServer) fetchRuleContent(rule types.RuleOnReport) (proxy_types.RuleWithContentResponse, bool) {
+// fetchRuleContent - fetching content for particular rule
+// Return values:
+//   - Structure with rules and content
+//   - return true if fetching content was successful, including filtering
+//   - return true if fetching and parsing of content was successful, not including filtering
+func (server HTTPServer) fetchRuleContent(rule types.RuleOnReport, OSDEligible bool) (proxy_types.RuleWithContentResponse, bool, bool) {
```
tisnik, line 467: at least the last `bool` seems it should be in fact `error`

server/server.go
Sergey1011010: but you could write an error earlier, when it happens in the loop and stop the whole thing, you ignore rules anyways

server/server.go
```diff
@@ -489,7 +499,7 @@ func (server HTTPServer) fetchRuleContent(rule types.RuleOnReport) (proxy_types.
 		Internal:        ruleWithContent.Internal,
 	}
 
-	return parsedRule, true
+	return parsedRule, true, true
```
Sergey1011010, line 502: Why 3rd one is true? According to the code below it is "isError", but it doesn't look like any error happened here.

Sergey1011010: also, having 2 return values indicating errors by opposite values(successful is true = "ok", isError is false = "ok") is kinda confusing

server/server.go
Sergey1011010: btw, do you really need to return an error if this condition `if OSDEligible && !ruleWithContent.NotRequireAdmin {` is met only for one rule? Maybe just ignore this rule

content/parsing.go
```diff
@@ -31,6 +32,15 @@ var (
...
 
+func contains(arr []string, str string) bool {
+	for _, a := range arr {
+		if a == str {
+			return true
+		}
+	}
+	return false
+}
+
```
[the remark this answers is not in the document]
bond95: Yes, we have, thank you

server/server.go
bond95: I'm not sure about that, because `fetchRuleContent` handling errors by itself, so in that case we just need to be sure if there is some errors.

server/server.go
bond95: Yes, and no, because if we handle errors in this place, will mean, that in logs we get all rules, that's couldn't be parsed or fetched, which is much more better than stop request after first error.

bond95: It's not returning error when it's OSD, it just return not successful state, which mean that this rule will be ignored.

server/server.go
bond95: But information about failure is handled before in inner functions.

server/server.go
bond95: Fixed variable name.

server/server.go
tisnik: ok agree. Could you then please name the return values? TYVM

content/parsing.go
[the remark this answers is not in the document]
tisnik: ok, thank you

server/server.go
bond95: @tisnik I already left a comment, is there enough, check it please?

tisnik: ok not a problem

tisnik (approved):
LGTM

server/server.go
Sergey1011010: Now you changed the logic and the error will be returned to the user if a single rule couldn't parse, not sure if we want this logic, but previous version was different(we just logged this error and returned what we could). @tisnik wdyt?

server/server.go
```diff
@@ -591,14 +620,19 @@ func (server HTTPServer) singleRuleEndpoint(writer http.ResponseWriter, request
 		return
 	}
 
-	rule, successful = server.fetchRuleContent(*aggregatorResponse)
+	rule, successful, _ = server.fetchRuleContent(*aggregatorResponse, server.getOSDFlag(request))
```
Sergey1011010, line 623: Why do you ignore error here?

server/server.go
```diff
@@ -591,14 +620,19 @@ func (server HTTPServer) singleRuleEndpoint(writer http.ResponseWriter, request
...
 
 	if !successful {
+		err := responses.SendNotFound(writer, "Rule was not found")
```
Sergey1011010, line 626: You could use only handleServerError here https://github.com/RedHatInsights/insights-operator-utils/blob/master/types/errors.go#L91

server/server.go
bond95: Because if at least something goes wrong then `successful` will be `false`, including errors. So it's no need to handle it.

server/server.go
bond95: `singleRuleEndpoint` don't get ruleID and ruleError, so using extracting that params second time just for this error, I think it's don't worth it https://github.com/RedHatInsights/insights-operator-utils/blob/15514d9c64e25edf80fee81e229b74c4b70b27e5/types/errors.go#L110

server/server.go
Sergey1011010: well, if somebody changes the implementation, then there will be a problem

server/server.go
Sergey1011010: `aggregatorResponse.Module`, but ok, not a big deal

joselsegura (approved):
LGTM

config.toml
```diff
@@ -3,8 +3,8 @@ address = ":8081"
 api_prefix = "/api/v1/"
 api_spec_file = "openapi.json"
 debug = true
-auth = false
-auth_type = "xrh"
+auth = true
+auth_type = "xhr"
```
quarckster, line 7: Why did you change it?

quarckster (changes requested):

config.toml
bond95: I'll change it back. I used it for local testing

bond95: Fixed

quarckster (approved):

PR #352, opened 2020-11-18
Returning 404 only in specific conditions

### LINKED ISSUE #351: 404 on GET /clusters/culsterId/report
insights-results-smart-proxy unable to show a report if at least one rule doesn't have content.

```
"status":"Item with ID [ccx_rules_ocp.external.rules.machineconfig_stuck_by_node_taints.report] was not found in the storage"}
```

Expected result.
Rules that missing the content should be excluded from the report and `/clusters/{culsterId}/report ` should return 200 and correct report for other rules.

joselsegura:
# Description

Some tuning to the condition where the report endpoint returns 404: only returning 404 if there are some hits that are missing content and there is not a single one rule hit that has its content.

Fixes #351 

## Type of change

## Testing steps

Regular CI. Working in adding some unit tests related to the issue


quarckster:
@bond95 could you please confirm that Insights tab will show the correct state for that corner case?

tisnik (approved):
LGTM. And I hope I will never see this code again :/

Bee-lee (approved):
LGTM

server/server.go
```diff
@@ -645,34 +645,16 @@ func (server HTTPServer) reportEndpoint(writer http.ResponseWriter, request *htt
...
-	var rules []proxy_types.RuleWithContentResponse
-	var rulesNotFound []types.RuleID
-	hasNoErrors := true
-
+	rules := []proxy_types.RuleWithContentResponse{}
 	for _, aggregatorRule := range aggregatorResponse.Report {
-		rule, successful, hasNoError := server.fetchRuleContent(aggregatorRule, server.getOSDFlag(request))
-
-		hasNoErrors = hasNoErrors && hasNoError
+		rule, successful, _ := server.fetchRuleContent(aggregatorRule, server.getOSDFlag(request))
```
bond95, line 650: I think you should refactor `fetchRuleContent` as well as it's not needed third parameter anymore

bond95 (changes requested):

bond95:
@quarckster according to what I see, only change that was done is, that `smart-proxy` will ignore any rule that doesn't have content or unable to fetch content. So maybe there is could be situation when map with content is empty, and it'll return empty array for any request.

server/server.go
tisnik: actually the refactoring must be deeper (I feel), IMHO should not be part of bug fix. WDYT?

joselsegura: thank you @bond95. I will create a separated issue for the refactor

bond95: @tisnik yep, must be, but removing third parameter is part of this PR, because you start unusing it here, so it should be removed here as well

tisnik (approved):
LGTM

server/server.go
```diff
@@ -645,43 +645,34 @@ func (server HTTPServer) reportEndpoint(writer http.ResponseWriter, request *htt
...
 			LastCheckedAt: aggregatorResponse.Meta.LastCheckedAt,
-			Count:         len(rules),
+			Count:         len(rules) + rulesWithoutContent,
 		},
 		Data: rules,
 	}
 
-	err := responses.SendOK(writer, responses.BuildOkResponseWithData("report", report))
+	status := http.StatusOK
+	if rulesWithoutContent > 0 {
```
bond95, line 671: We talked about if 100% of rules will be without content no?
Your quote:
```
N hits: at least one with content: status 200, report length: 1, meta count: N
N hits: 0 with content: 500, report length 0 and meta count N ?
```

server/server.go
```diff
@@ -668,7 +668,7 @@ func (server HTTPServer) reportEndpoint(writer http.ResponseWriter, request *htt
 	}
 
 	status := http.StatusOK
-	if rulesWithoutContent > 0 {
+	if rulesWithoutContent > 0 && len(rules) == 0 {
```
tisnik, line 671: :+1:

bond95 (approved):

quarckster (approved):

PR #454, opened 2021-04-05
Do not send rules reported that are disabled for the cluster

joselsegura:
# Description

The rules included for each report in the endpoint for a single cluster includes the disabled rules too. The report should not include those rules.

Fixes #[insights-results-aggregator#1128](https://github.com/RedHatInsights/insights-results-aggregator/issues/1128)

## Type of change

## Testing steps

Regular CI

## Checklist


tisnik (approved):
Is it THAT easy? :) LGTM of course

joselsegura:
Well, I need to double check because I think we are still storing the "disabled" flag and user votes per user and not per cluster, so I want to discuss it tomorrow when the crew is back to work :-)

Bee-lee (changes requested):
If we merge this, we would lose the functionality in OCM to re-enable the rule.
The `disabled` flag is checked on OCM frontend and such disabled rules are then put in the disabled health checks section, from which the user has the ability to re-enable the rule.

joselsegura:
> If we merge this, we would lose the functionality in OCM to re-enable the rule.
> The `disabled` flag is checked on OCM frontend and such disabled rules are then put in the disabled health checks section, from which the user has the ability to re-enable the rule.

Very good point @Bee-lee I didn't think on the OCM part at all. I will review with somebody from presentation team to be sure to don't broke anything

Bee-lee (approved):
LGTM!

dpensi (approved):

server/server.go
```diff
@@ -825,3 +772,34 @@ func (server HTTPServer) getOverviewPerCluster(
...
+) {
+	filteredRules = []proxy_types.RuleWithContentResponse{}
+	noContentRules = 0
+
+	for _, aggregatorRule := range aggregatorReport {
+		if aggregatorRule.Disabled && filterDisabled {
+			continue
+		}
+
+		rule, successful, filtered := content.FetchRuleContent(aggregatorRule, filterOSD)
```
tisnik, line 792: are we blindly "forget" rules w/o content there?

server/server_test.go
```diff
@@ -36,7 +36,7 @@ import (
 )
 
 const (
-	testTimeout            = 10 * time.Second
+	testTimeout            = 5 * time.Second
```
tisnik, line 39: it won't work in all times on CI, really ;)

tisnik (approved):
LGTM

server/server.go
joselsegura: The next `if` block checks it and increase the `noContentRules` variable, which is one of the returned ones. If I'm not mistaken, the code behaves exactly as the previous version (which probably needs a refactor, maybe as a different task)

server/server_test.go
joselsegura: Ouch, I forgot to increase it again. I'll remove this change

PR #471, opened 2021-04-22
Fixing problem when there is an empty result because disabled rules

### LINKED ISSUE #470: Wrong status code returned in specific scenarios
Given the following scenario:

1. Cluster that has, at least, 2 rule hits
2. All the rules are disabled
3. There is at least 1 rule that cannot be shown because it lacks of content in the content service

Then, the response contains the expected report, but the status code is 500 instead of 200

joselsegura:
# Description

This fix takes into account the number of disabled rules to decide when send an error status or not

Fixes #470 

## Type of change

## Testing steps

Regular CI

## Checklist


Bee-lee (approved):
TY!

tisnik (approved):
LGTM

PR #487, opened 2021-05-12
Exclude some rules due to `pr_check` script

tisnik:
# Description

Exclude some rules due to `pr_check` script

## Type of change

## Testing steps

Done in CI

## Checklist


quarckster (approved):

epapbak (approved):
Some of these rules are easily to comply with. Maybe we should open an issue for fixing the script?

PR #561, opened 2021-07-16
Update feedback endpoints

joselsegura:
# Description

This PR updates the endpoints used for voting, enabling, disabling and giving feedback for the rule hits from an Smart Proxy client. Now, it will require the "error key", not only the rule id, because some of the rules have more than 1 valid error key

## Type of change

## Testing steps

Regular CI

## Checklist


tisnik (approved):
LGTM

PR #568, opened 2021-07-29
make status and publish_date optional but disallow GIGO

Bee-lee:
# Description

This PR makes the `status` and `publish_date` attributes of error keys optional, but does not allow garbage in, garbage out.
With the checks performed by Martin's script, we should be safe.

Fixes CCXDEV-5442

## Type of change

## Testing steps
smart-proxy gets 141 rules from content-service and with these changes, smart-proxy provides content for all 141 rules, `make before_commit` and also checked visually if there is no garbage in the current content

## Checklist

### DIFF
--- a/content/parsing.go
+++ b/content/parsing.go
@@ -118,20 +118,20 @@ func commaSeparatedStrToTags(str string) []string {
 	return strings.Split(str, ",")
 }
 
-func timeParse(value string) (time.Time, bool, error) {
-	var err error
-	missing := false
+func timeParse(value string) (publishDate time.Time, missing bool, err error) {
+	missing = false
+	publishDate = time.Time{}
 
 	if value == "" {
 		missing = true
-		return time.Time{}, missing, nil
+		return
 	}
 
 	for _, datetimeLayout := range timeParseFormats {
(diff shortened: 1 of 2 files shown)


content/parsing.go
```diff
@@ -111,13 +118,20 @@ func commaSeparatedStrToTags(str string) []string {
...
 }
 
-func timeParse(value string) (time.Time, error) {
+func timeParse(value string) (time.Time, bool, error) {
 	var err error
+	missing := false
+
+	if value == "" {
+		missing = true
+		return time.Time{}, missing, nil
```
tisnik, line 127: probably good idea to use named return values (for more than two return values)

tisnik (approved):
LGTM

content/parsing.go
Bee-lee: it rarely makes the code more readable for metime, I only use it when the return values are of the same type, like here in the getActiveStatus(), which was bool, bool, bool.. But let's keep it consistent, it's a good idea

dpensi (approved):

joselsegura (approved):
lgtm

PR #570, opened 2021-08-09
Extend log messages for edge case scenario

Bee-lee:
# Description

So.... you see this condition that I've been ignoring since I started looking into this, because I read it wrong the first time and made myself completely sure it's just a sanity check for an impossible scenario that could never happen -- hence the Internal Server Error, right? These totally obvious lines of code that literally have the words `InternalServerError` in them? Yup! Today I realized that I might've been chasing a ghost this whole time :smiling_face_with_tear: 

If this is it, then I really learned not to trust myself and double check everything .... :) 

## Checklist


tisnik (approved):
Nice one, LGTM

joselsegura (approved):
LGTM

PR #572, opened 2021-08-10
No content when rules are hitting is no longer causing 500s

Bee-lee:
# Description
I totally ignored the if statement when I started working on it because I misread it and considered it an impossible scenario "sanity" check... 

This case should appear as "No issues found" in customer-facing applications, because the only we could show is rule module + error key, which have no informational value to customers, so now this edge case will appear as if the cluster had no issues in OCM, which is actually true from the customer's POV, since we can't even describe the problem, let alone provide resolution.

Fixes (hopefully) https://bugzilla.redhat.com/show_bug.cgi?id=1977858

## Type of change

## Testing steps
`make before_commit`, two unit test cases can now be considered as reproducers

## Checklist

### DIFF
--- a/server/handlers_test.go
+++ b/server/handlers_test.go
@@ -39,6 +39,15 @@ var (
 			Impact: testdata.ImpactStrToInt,
 		},
 		Rules: map[string]types.RuleContent{
+			"rc1": testdata.RuleContent1,
+		},
+	}
+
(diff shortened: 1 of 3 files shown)


tisnik (approved):
Very nice one, indeed!

joselsegura (approved):

PR #583, opened 2021-08-20
Fixed smart-proxy service name

quarckster:
# Description

We should keep the old service name to have 3scale working.

## Type of change

### DIFF
--- a/deploy/clowdapp.yaml
+++ b/deploy/clowdapp.yaml
@@ -24,24 +24,24 @@ objects:
   metadata:
     labels:
       app: ccx-data-pipeline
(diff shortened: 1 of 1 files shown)


joselsegura (approved):
TBH it looks very weird that the "name" is "ccx-smart" and deployment name only "proxy".

If it is needed, maybe it could be better to separate them as "ccx" and "smart-proxy", if possible

quarckster:
@joselsegura done

joselsegura (approved):

tisnik (approved):
LGTM

PR #584, opened 2021-08-20
Use POST endpoint instead of expand the cluster list

joselsegura:
# Description

`org_overview` endpoint currently used the GET endpoint to retrieve all the cluster reports from Aggregator. It can lead in some situations to very long URLs that can be handled differently depending on configuration.

This change makes the endpoint to use the POST endpoint version to retrieve the reports for a list of clusters, which will avoid the possible URL problems

## Type of change

## Testing steps

Regular CI

## Checklist

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -429,16 +429,23 @@ func (server HTTPServer) readAggregatorReportForClusterListFromBody(
 		return nil, false
 	}
 
+	return handleReportsResponse(aggregatorResp, writer)
+}
+
+// handleReportsResponse analyses the aggregator's response and
+// writes an appropriate response to the client, handling any
+// possible error in the meantime
+func handleReportsResponse(response *http.Response, writer http.ResponseWriter) (*types.ClusterReports, bool) {
 	var aggregatorResponse types.ClusterReports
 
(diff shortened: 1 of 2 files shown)


dpensi (approved):

juandspy (approved):

server/server.go
```diff
@@ -407,6 +408,53 @@ func (server HTTPServer) readAggregatorReportForClusterList(
...
+		handleServerError(writer, err)
+		return nil, false
+	}
+	// #nosec G107
+	aggregatorResp, err := http.Post(aggregatorURL, "application/json", bytes.NewBuffer(body))
+	if err != nil {
+		handleServerError(writer, err)
+		return nil, false
+	}
+
```
tisnik, line 431: probably will be better to split the function at this line. The reason - free the `body` + associated `Buffer` as fast as possible as it is memory intensive.

tisnik (commented):
One comment, LGTM otherwise.

server/server.go
joselsegura: Makes sense. Updating

tisnik (approved):
LGTM

PR #606, opened 2021-09-10
Adding AMS API client to Smart Proxy and org_overview

joselsegura:
# Description

Adding a client to interact with AMS API.
Modified org_overview endpoint handler to use this client if available

WIP: lack of unit tests for the `amsclient` package, probably too low coverage

## Type of change

## Testing steps

Regular CI

## Checklist
* [] updated schemas and validators in [insights-data-schemas](https://github.com/RedHatInsights/insights-data-schemas) in case of input/output change


amsclient/amsclient.go
```diff
@@ -0,0 +1,131 @@
...
+		Build()
+
+	if err != nil {
+		log.Error().Err(err).Msg("Unable to build the connection to AMS API")
+		return nil, err
+	}
+
+	var pageSize int
+	if conf.PageSize == 0 {
+		pageSize = 100
```
tisnik, line 47: prob.worth to add constant?

amsclient/amsclient.go
```diff
@@ -0,0 +1,131 @@
...
+	var retval []types.ClusterName = []types.ClusterName{}
+
+	internalOrgID, err := c.getInternalOrgIDFromExternal(orgID)
+	if err != nil {
+		return retval
+	}
+
+	searchQuery := fmt.Sprintf("organization_id is '%s'", internalOrgID)
+
+	if len(statusFilter) > 0 {
```
tisnik, line 70: I'd refactor it into a new function

amsclient/amsclient.go
```diff
@@ -0,0 +1,131 @@
...
+		clusterIDQuery := " and status in ('" + strings.Join(statusFilter, "',") + "')"
+		searchQuery = searchQuery + clusterIDQuery
+	}
+
+	if len(statusNegativeFilter) > 0 {
```
tisnik, line 75: dtto

amsclient/amsclient.go
```diff
@@ -0,0 +1,131 @@
...
+		for _, item := range response.Items().Slice() {
+			clusterID, ok := item.GetExternalClusterID()
+			if !ok {
+				fmt.Println("Not external cluster ID")
+				continue
+			}
+			retval = append(retval, types.ClusterName(clusterID))
+		}
+
+		pageNum++
```
tisnik, line 105: hmm the `pageNum` can be handled in the `for` loop, or not?

amsclient/configuration.go
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package amsclient
+
+// Configuration represents the configuration of the AMS API client
+type Configuration struct {
+	Token    string `mapstructure:"token" toml:"token"`
+	URL      string `mapstructure:"url" toml:"url"`
+	PageSize int    `mapstrictire:"page_size" toml:"page_size"`
```
tisnik, line 21: +default page size perhaps? (your `100` constant above)

tisnik (commented):
LGTM in overall, just some nitpicks :)

amsclient/configuration.go
joselsegura: Is it something that can be done in annotations? I will search for it, but I agree that, at least, I need to define a const for the default value

amsclient/amsclient.go
joselsegura: Agree, moving the filter string generation to an utility function

amsclient/amsclient.go
joselsegura: Yes, I can. The only think that I cannot put in the `for` loop to be handled is the exit condition, as it depends on the response.

amsclient/amsclient.go
```diff
@@ -0,0 +1,134 @@
...
+
+		// When an empty page is returned, then exit the loop
+		if response.Size() == 0 {
+			break
+		}
+
+		for _, item := range response.Items().Slice() {
+			clusterID, ok := item.GetExternalClusterID()
+			if !ok {
+				fmt.Println("Not external cluster ID")
```
tisnik, line 99: prob. log would be better

amsclient/amsclient.go
```diff
@@ -0,0 +1,134 @@
...
+		Fields("id,external_id").
+		Send()
+
+	if err != nil {
+		log.Error().Err(err).Msg("")
+		return "", err
+	}
+
+	if response.Items().Len() != 1 {
+		log.Error().Int("orgIDs length", response.Items().Len()).Msg("More than one organization for the given orgID")
```
tisnik, line 123: could you please add org_id into the log message too?

amsclient/amsclient.go
```diff
@@ -0,0 +1,134 @@
...
+		return "", fmt.Errorf("More than one organization for the given orgID (%d)", orgID)
+	}
+
+	internalID, ok := response.Items().Get(0).GetID()
+	if !ok {
+		log.Error().Msgf("Organization %d doesn't have proper internal ID", orgID)
+		return "", fmt.Errorf("Organization %d doesn't have proper internal ID", orgID)
```
tisnik, line 130: formatting done twice, probably `goconst` won't be too happy

server/server.go
```diff
@@ -61,6 +62,7 @@ type HTTPServer struct {
 	Config         Configuration
 	ServicesConfig services.Configuration
 	GroupsChannel  chan []groups.Group
+	amsClient      *amsclient.AMSClient
```
tisnik, line 65: is it lowercase by mistake?

server/server.go
```diff
@@ -301,6 +316,10 @@ func copyHeader(srcHeaders http.Header, dstHeaders http.Header) {
 // readClusterIDsForOrgID reads the list of clusters for a given
 // organization from aggregator
 func (server HTTPServer) readClusterIDsForOrgID(orgID types.OrgID) ([]types.ClusterName, error) {
+	if server.amsClient != nil {
+		return server.amsClient.GetClustersForOrganization(orgID, nil, []string{"Deprovisioned", "Archived"}), nil
```
tisnik, line 320: we are gonna to use that code multiple times, so what about making those two strings constants?

tisnik (approved):
LGTM, just couple of nitpicks.

Also: if page size would be `uint` it might be more correct (but really low prio)

Bee-lee (approved):
looks really good!

server/server.go
joselsegura: No, it is just to don't expose it, as it shouldn't be used directly by any uer of the HTTPServer objects. It can be changed, of course

server/server.go
joselsegura: Agree

amsclient/amsclient.go
joselsegura: Ok, I'll extract it to some const out there

amsclient/amsclient.go
joselsegura: Sure!

amsclient/amsclient.go
joselsegura: Agree, I forgot to change this one

PR #651, opened 2021-10-01
Use internal tools for mocking instead Gock requests

### LINKED ISSUE #650: Unify content mocking in server_test
`server_test` package contains a lot of tests for endpoints that need mocked content to be loaded in order to work.

Right now there are 2 different mechanisms used:

- Using `gock` library to mock the real requests performed by Smart Proxy to Content Service, faking the response
- Using internal utils in `content` package for loading mocked data

The first one has a very important problem, because it starts the "content update loop", which can lead on race conditions between tests (tests passing when run individually, but failing when running the whole test suite)

The second one is much faster, doesn't depend on external dependency and decrease the chances to a race condition to happen

Chose one and go for it

joselsegura:
# Description

From some time ago, the unit tests are suffering from race conditions, unwanted waiting times and difficult content mocking for develop our endpoint tests. This PR tries to reduce that technical debt and increase the speed when running our unit tests (very annoying issue when trying locally)

Fixes #650 

## Type of change

## Testing steps

Regular CI

## Checklist

### DIFF
--- a/server/handlers_test.go
+++ b/server/handlers_test.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"net/http"
 	"testing"
-	"time"
 
 	"github.com/RedHatInsights/insights-operator-utils/responses"
 	"github.com/RedHatInsights/insights-operator-utils/types"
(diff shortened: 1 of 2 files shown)


tisnik (approved):
LGTM, it's much more unit-test-ish

dpensi (approved):

PR #653, opened 2021-10-01
Dependent services unavailable error code

### LINKED ISSUE #652: Endpoints should return HTTP status 503 in case content service/aggregator is not reachable
Currently, smart proxy endpoints return default status 500 in case some of the dependent services is unavailable. 503 should be more suitable for that purpose. More info [here](https://issues.redhat.com/browse/CCXDEV-5694).

JiriPapousek:
# Description

All endpoints should now return 503 in case that either content service or aggregator are unavailable. Some major modifications in the code were needed.

Fixes #652 

## Type of change

## Testing steps

Each of the endpoints has been manually checked, both with functioning depending services and without them.

## Checklist


content/content.go
```diff
@@ -1,6 +1,6 @@
 // Copyright 2020 Red Hat, Inc
 //
-// Licensed under the Apache License, Version 2.0 (the "License");
+// Licensed under the Apac License, Version 2.0 (the "License");
```
tisnik, line 3: Apac? :)

content/content.go
```diff
@@ -140,15 +140,37 @@ var rulesWithContentStorage = RulesWithContentStorage{
...
+
+		done := make(chan struct{})
+		go func() {
+			ruleContentDirectoryReady.Wait()
+			close(done)
+		}()
+
+		select {
+		case <-done:
+		case <-time.After(5 * time.Second):
```
tisnik, line 164: prob worth to make 5 named constant?

tisnik (approved):
Looks ok, just few nitpicks

tisnik:
And please rebase first :)

content/content.go
JiriPapousek: Fixed

content/content.go
JiriPapousek: Fixed

tisnik:
@JiriPapousek LGTM, btw the fix for CI issue is easy. Just `Squash and merge` after you fix the style:)

content/content.go
```diff
@@ -140,15 +141,38 @@ var rulesWithContentStorage = RulesWithContentStorage{
...
+func WaitForContentDirectoryToBeReady() error {
 	// according to the example in the official dock,
 	// lock is required here
 	if ruleContentDirectory == nil {
 		ruleContentDirectoryReady.L.Lock()
-		ruleContentDirectoryReady.Wait()
+
+		done := make(chan struct{})
+		go func() {
+			ruleContentDirectoryReady.Wait()
```
dpensi, line 160: Hi! I tried to add a print at the end of this function, `log.Info().Msg("Unlocking the lock")` at line 171, but it never get printed and the endpoints don't seem to work , what is it waiting for here?

Edit: approving, probably i just don't understand how to test it

smart_proxy.go
```diff
@@ -148,16 +167,48 @@ func updateGroupInfo(servicesConf services.Configuration, groupsChannel chan []g
...
+	errorFoundChannel chan bool,
+	errorChannel chan error) {
+	for len(errorFoundChannel) > 0 {
+		<-errorFoundChannel
+	}
+	for len(errorChannel) > 0 {
+		<-errorChannel
+	}
+	for len(groupsChannel) > 0 {
+		<-groupsChannel
```
Bee-lee, line 208: LGTM overall, just one question, I see you initialize `errorFoundChannel` and `errorChannel` with size=1, but `groupsChannel` isn't, right? I don't remember how the groupsChannel works completely, but is it enough to empty it like this?

server/server.go
```diff
@@ -746,10 +804,16 @@ func (server HTTPServer) singleRuleEndpoint(writer http.ResponseWriter, request
 	if err != nil {
 		log.Err(err).Msgf("Got error while parsing `%s` value", OSDEligibleParam)
 	}
-	rule, successful, _ = content.FetchRuleContent(*aggregatorResponse, osdFlag)
+	rule, successful, _, err = content.FetchRuleContent(*aggregatorResponse, osdFlag)
```
epapbak, line 807: We can probably have error only, instead of successfull and error. WDYT?

PR #656, opened 2021-10-01
add smart-proxy part for /rule/ recommendation list

Bee-lee:
# Description

GET /rule/ which returns a list of recommendations for the active clusters in the organization (given the AMS API is alive) with a number the number of clusters it's impacting.

- because the filtering and sorting will be performed on the frontend for now, I tried to optimzie it as best as I could
- adds new variables to rule content storage to optimize processing required for new /rule/ related endpoints

## Type of change

## Checklist


content/content.go
```diff
@@ -76,6 +89,17 @@ func (s *RulesWithContentStorage) GetRuleContent(ruleID types.RuleID) (*types.Ru
 	return res, found
 }
 
+// GetContentForRecommendation returns content for rule with error key
+func (s *RulesWithContentStorage) GetContentForRecommendation(
+	ruleID types.RuleID,
+) (*local_types.RuleWithContent, bool) {
+	s.RLock()
+	defer s.RUnlock()
```
tisnik, line 97: I'm not 100%, but won't it be better to use `sync.Map`? https://golangforall.com/en/post/sync-map.html

content/content.go
```diff
@@ -93,13 +117,27 @@ func (s *RulesWithContentStorage) GetAllContent() []types.RuleContent {
...
+
 	s.Lock()
 	defer s.Unlock()
 
 	s.rulesWithContent[ruleIDAndErrorKey{
 		RuleID:   ruleID,
 		ErrorKey: errorKey,
 	}] = ruleWithContent
+
+	switch ruleWithContent.Internal {
```
tisnik, line 135: if-else? :)

content/content.go
```diff
@@ -117,11 +155,14 @@ func (s *RulesWithContentStorage) ResetContent() {
 	s.Lock()
 	defer s.Unlock()
 
-	s.rulesWithContent = make(map[ruleIDAndErrorKey]*local_types.RuleWithContent)
 	s.rules = make(map[types.RuleID]*types.RuleContent)
+	s.rulesWithContent = make(map[ruleIDAndErrorKey]*local_types.RuleWithContent)
+	s.recommendationsWithContent = make(map[types.RuleID]*local_types.RuleWithContent)
+	s.internalRuleIDs = make([]types.RuleID, 0)
```
tisnik, line 161: actually you can leave it as `nil` as the first `assert` will create a slice for you (nitpick)

PR #679, opened 2021-10-19
change behaviour of impacting param on GET /rule/

Bee-lee:
# Description
If param is missing, endpoint will return all available rules including those that don't hit any clusters at the moment. If set to true, only returns impacting rules. If false, only returns those that aren't imapcting."

Fixes CCXDEV-6177

## Type of change

- REST API tests

## Testing steps

`make before_commit` + local

## Checklist


tisnik (approved):
LGTM

dpensi (approved):

PR #681, opened 2021-10-19
fixed single rule endpoint

dpensi:
# Description

fixed single rule endpoint for api v1 and v2

Fixes api v1 fix

## Type of change

## Checklist


joselsegura (approved):
seems legit

PR #703, opened 2021-11-02
add more specific log message to readClusterIDsForOrgID

Bee-lee:
# Description

I was trying to find the `amsclient is not able to retrieve cluster list. Using fallback method`, which wasn't to be found in logs, so I assumed it was using the AMS API correctly, but I didn't notice that that part is only ran if `amsClient` is initialized, which isn't the case on QA I suppose.

## Testing steps
n/a

## Checklist

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -372,7 +372,9 @@ func (server HTTPServer) readClusterIDsForOrgID(orgID types.OrgID) ([]types.Clus
 			return clusters, err
 		}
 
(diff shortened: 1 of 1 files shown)


dpensi (approved):

tisnik (approved):
LGTM, of course

joselsegura:
LGTM of course. The `amsclient` package prints a message if it is not initialised, but only at the very beginning of the start up of the service

PR #706, opened 2021-11-03
Not using functions from Go 1.16, base image cannot be upgraded yet

### LINKED ISSUE #705: Container image cannot be built
Last tries of building the service image (https://ci.ext.devshift.net/job/RedHatInsights-insights-results-smart-proxy-gh-build-master/122/console) are failing because `io.ReadAll` is not defined in Go 1.15, that is still being used as base image.

joselsegura:
# Description

Right now the conainter image build of Smart Proxy are failing because Go 1.15 doesn't have `io.ReadAll` function (added in 1.16). As long as we cannot use go-toolset:1.16 version yet, we should stick to 1.15 version.

Downgrading Travis CI version to 1.15 too

Fixes #705

## Type of change

## Testing steps

Container image built locally without any problem

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,7 @@
 
 language: go
 go:
-- "1.15"
-
+- "1.16"
 jobs:
   include:
     - stage: style
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@
 
 language: go
 go:
-- 1.16
+- "1.15"
 
 jobs:
   include:
--- a/server/handlers_v1.go
+++ b/server/handlers_v1.go
@@ -18,7 +18,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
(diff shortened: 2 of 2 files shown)


tisnik (approved):
ok

juandspy (approved):

.travis.yml
```diff
@@ -14,8 +14,7 @@
 
 language: go
 go:
-- 1.16
-
+- "1.16"
```
tisnik, line 17: should it be in quotes?

dpensi (approved):

.travis.yml
joselsegura: Yes! I found in Travis documentation. It is specially important when referring to a 0-ended version. The example was that, for Go 1.10, if you don't use quotes, you will get the 1.1 version instead of the desired 1.10, because it is considered as a float number :-)

PR #709, opened 2021-11-04
Return empty list and 200OK when no impacted clusters found for given recommendation

epapbak:
# Description

- With this change, the recommendation provided to the `rule/{rule_selector}/clusters_detail` endpoint is first checked for existence, and in case it is not in the existing external rules, a 404 is returned.
- If the rule exists and the DB returns an ItemNotFound error, the proxy will return a JSON with an empty `data` field and a `200 OK` status instead of proxying the 404 error to the client. 

## Type of change

## Testing steps

Updated UTs

## Checklist

### DIFF
--- a/server/handlers_v2.go
+++ b/server/handlers_v2.go
@@ -454,9 +454,9 @@ func (server HTTPServer) getImpactedClusters(
 	)
 
 	var aggregatorResp *http.Response = nil
+	var err error
 
 	if len(activeClusters) < 0 {
-		var err error
 		// #nosec G107
 		aggregatorResp, err = http.Get(aggregatorURL)
 		// if http.Get fails for whatever reason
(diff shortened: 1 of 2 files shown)


server/handlers_v2.go
```diff
@@ -466,16 +466,17 @@ func (server HTTPServer) getImpactedClusters(
 		}
 	} else {
 		// generate JSON payload of the format "clusters": []clusters
-		jsonBody, err := json.Marshal(
+		jsonBody, e := json.Marshal(
```
juandspy, line 469: why not `err`?

juandspy (approved):

server/handlers_v2.go
```diff
@@ -454,9 +454,9 @@ func (server HTTPServer) getImpactedClusters(
 	)
 
 	var aggregatorResp *http.Response = nil
+	var err error
```
tisnik, line 457: This function is huge, with many `err`s having different meaning etc. Probably it is a right time to refactor it. Another rule of thumb - is anybody able to describe the function in one simple English sentence?

tisnik (changes requested):
Probably worth to spend some time on refactoring?

server/handlers_v2.go
epapbak: because the linter that detects ineffectual assignments complains when I reuse err everywhere, and I don't want to fight with it =)

server/handlers_v2.go
epapbak: cleaned it up a little bit

tisnik (approved):
LGTM

server/handlers_v2.go
epapbak: it looks better post-refactor

PR #717, opened 2021-11-09
incerase AMS API page size to 1500 as the separate requests still take longer

Bee-lee:
# Description
let's find the soft limit of AMS API because making more requests currently seems slower than fewer but bigger requests. Currently over 90% of time is spent talking with AMS API with an organization with >4400 clusters. The rest, even inclduing talking with aggregator takes just a couple of tenths

## Type of change

## Testing steps
n/a

## Checklist


tisnik (approved):
Let's DOS it!

juandspy (approved):

PR #719, opened 2021-11-11
Make AMS API fallback mechanism configurable

joselsegura:
# Description

With this PR the fallback mechanism in case that AMS API is not configured or it isn't working is configurable.

Refactored the amsclient and its usages to make it easier to use a test double.
Refactored unit tests that rely on the fallback mechanism
Added new UT for testing fallback mechanism in case it is needed

## Type of change

## Testing steps

Tested though unit tests

## Checklist

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -74,14 +74,15 @@ const (
 
 // HTTPServer is an implementation of Server interface
 type HTTPServer struct {
-	Config            Configuration
-	InfoParams        map[string]string
-	ServicesConfig    services.Configuration
-	amsClient         amsclient.AMSClient
-	GroupsChannel     chan []groups.Group
-	ErrorFoundChannel chan bool
-	ErrorChannel      chan error
-	Serv              *http.Server
+	Config                 Configuration
+	InfoParams             map[string]string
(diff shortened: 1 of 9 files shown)


dpensi (approved):

Bee-lee (approved):
looks good, pretty elegant 
so just calling `readClusterIDsForOrgID` will take care of it, right?

server/server.go
```diff
@@ -74,14 +74,15 @@ const (
...
-	ErrorChannel      chan error
-	Serv              *http.Server
+	Config                 Configuration
+	InfoParams             map[string]string
+	ServicesConfig         services.Configuration
+	amsClient              amsclient.AMSClient
+	GroupsChannel          chan []groups.Group
+	ErrorFoundChannel      chan bool
+	ErrorChannel           chan error
+	useOrgClustersFallback bool
```
tisnik, line 84: ah should not it be part of `Config`?

joselsegura: Mmm, and in fact, it is! It is really duplicated. I will push a new commit. Good catch!

tisnik (approved):
ok

PR #720, opened 2021-11-12
Adding display name to the results gathered by amsclient

joselsegura:
# Description

Adding "display_name" column from AMS API into the results generated by the `amsclient` package.
Adapted `amsclient` unittests
Adapted `server` package code relying on `amsclient` output
New type for `ClusterInfo`, just in case we need more things apart of ID and display name in the future
Adapted `server` unit tests, as they use the mocked `amsclient` version

## Type of change

## Testing steps

Tested locally against stage environment and unit tests

## Checklist

### DIFF
--- a/amsclient/amsclient.go
+++ b/amsclient/amsclient.go
@@ -22,7 +22,7 @@ import (
 	sdk "github.com/openshift-online/ocm-sdk-go"
 	"github.com/rs/zerolog/log"
 
-	"github.com/RedHatInsights/insights-results-aggregator/types"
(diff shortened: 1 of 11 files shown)


tisnik (approved):
OK. Dunno about the type aliases, but it seems legit

joselsegura:
> OK. Dunno about the type aliases, but it seems legit

I did it just for avoiding using 2 different imports in some places, probably they are not needed at the end

joselsegura:
I forgot to add one new file to the commit, sorry for that

dpensi (approved):

juandspy (approved):

PR #724, opened 2021-11-25
[CCXDEV-6254] use testdata RuleContent1V2

juandspy:
# Description

Use https://github.com/RedHatInsights/insights-results-aggregator-data/pull/170 to resolve the testing problems we were having due to V2 rule types.

## Type of change

## Testing steps

Unit tests.

## Checklist

### DIFF
--- a/content/content_test.go
+++ b/content/content_test.go
@@ -59,7 +59,7 @@ func TestGetRuleContent(t *testing.T) {
 		helpers.FailOnError(t, err)
 		assert.NotNil(t, ruleContent)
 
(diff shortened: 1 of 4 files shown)


tisnik (approved):
ok, but it is not enough -> should update go.mod, etc.

juandspy:
> ok, but it is not enough -> should update go.mod, etc.

@tisnik already updated. Maybe after you checked the code. It should work now. At least, `make before_commit` passes in local.

dpensi (approved):

PR #727, opened 2021-11-26
Return cluster names in clusters details

joselsegura:
# Description

Adding the "cluster_name" in the response of the cluster_details endpoint, added to the existing cluster_id.

## Type of change

## Testing steps

Testes though local environment and unit tests

## Checklist

### DIFF
--- a/server/handlers_v2.go
+++ b/server/handlers_v2.go
@@ -519,6 +519,7 @@ func proxyImpactedClusters(
 	writer http.ResponseWriter,
 	selector ctypes.RuleSelector,
 	aggregatorResp *http.Response,
+	namesMap map[ctypes.ClusterName]string,
 ) error {
 	// If we received a 404 - no entries found for given orgID+selector in DB
 	// We return an empty list and a 200 OK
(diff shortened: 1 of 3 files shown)


juandspy (approved):

Bee-lee:
please also update OpenAPI spec

joselsegura:
> please also update OpenAPI spec

It was already updated: https://github.com/RedHatInsights/insights-results-smart-proxy/blob/8de6370044c9d330369f35b3a1c874ee3b9b2f31/server/api/v2/openapi.json#L841

This is because the internal data struct used by the response was updated already, so the response was including the clustern_name, but it will be empty until this PR is merged

Bee-lee (commented):
looks ok, are we sure this is not a breaking change for the UI? I know its still a list of structs with an extra field, but maybe it would be worth to check? esp. considering it's already used in prod beta

dpensi (approved):

dpensi (approved):

tisnik (commented):
Does it change the REST API?

joselsegura:
> looks ok, are we sure this is not a breaking change for the UI? I know its still a list of structs with an extra field, but maybe it would be worth to check? esp. considering it's already used in prod beta

I don't have experience with the Advisor UI. How can I check it? In addition, which backend environment is used by prod beta?

joselsegura:
> Does it change the REST API?

No, it doesn't. The API is already returning the field, but always with empty value

tisnik (approved):
ok

tisnik:
> > Does it change the REST API?
> 
> No, it doesn't. The API is already returning the field, but always with empty value

ok then :)

do you plan to add some REST API test?

joselsegura:
> > > Does it change the REST API?
> > 
> > 
> > No, it doesn't. The API is already returning the field, but always with empty value
> 
> ok then :)
> 
> do you plan to add some REST API test?

I didn plan it yet, I will create a task for it (I will need to create the whole rest_api_tests skeleton because it doesn't exist for this repo)

PR #830, opened 2021-12-08
extend log messages for duplicate cluster IDs from AMS API

Bee-lee:
# Description
there is something really wrong, the clusterNamesMap has fewer elements than the list -- there are probably duplicate cluster IDs... let's see

## Testing steps
n/a

## Checklist


tisnik (approved):
LGTM. btw interesting how it should happen :)

dpensi (approved):

PR #831, opened 2021-12-08
fix AMS API not enforcing unique or filled in cluster UUIDs

Bee-lee:
# Description
Apparently, AMS API is not enforcing unique cluster UUIDs. We were storing the cluster display names in a map with cluster UUIDs as keys, but the map had fewer elements than the list. After some investigation, the AMS API is not only not enforcing uniqueness, but they allow totally EMPTY UUIDs.

Fixes CCXDEV-6650

## Type of change

## Testing steps
make test

## Checklist


tisnik (approved):
ok

dpensi (approved):

PR #834, opened 2021-12-09
change AMS API query to match query OCM makes

Bee-lee:
# Description
This PR changes the AMS client to make the same query to get the list of clusters as OCM.
We were missing an extra filter `cluster_id != ''` which meant we were showing a few more clusters than OCM does.

Fixes CCXDEV-6797

## Type of change

## Testing steps
`make test`

## Checklist


tisnik (approved):
LGTM
(OTOH for future - probably it would be worth to perform URL encoding programatically)

joselsegura (approved):

PR #860, opened 2022-01-31
Org overview v1 acked rules

epapbak:
# Description

Take into account ack'ed rules in `org_overview` endpoint (v1 API)

Fixes CCXDEV-7097

## Type of change

- Unit tests

## Testing steps

Added new UTs and adapted existing ones to test the new behaviour

## Checklist

### DIFF
--- a/server/acks_handlers.go
+++ b/server/acks_handlers.go
@@ -18,6 +18,7 @@ package server
 
 import (
 	"encoding/json"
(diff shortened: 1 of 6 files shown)


tisnik (approved):
Seems ok (OTOH I still think that the 'map' should be a 'set' :)

joselsegura (approved):

PR #865, opened 2022-02-04
Remove .report substring from rule_selector used in v2 content endpoints

epapbak:
# Description

UI will start sending the rule_selector as it comes in the reports. Therefore, the module part will contain ".report", which our v2-related tables do not expect. Additional considerations:
- This is a quick fix. We have a readAndTrimRuleSelector method in insights-operator-utils that we can't use here because the parameter expected in the URL is `rule_id` instead of `rule_selector`. Once the fix is tested and approved, we should refactor the URLs and use only one method (and update openAPI spec in this repo and for integration tests)
- This change is not breaking, since if there is no `.report|` in the recommendation, nothing is done.

Fixes [CCXDEV-7183](https://issues.redhat.com/browse/CCXDEV-7183)

## Type of change

## Testing steps

Please describe how the change was tested locally. If, for some reason, the testing was not done or not done fully, please describe what are the testing steps.

## Checklist

### DIFF
--- a/server/router_utils.go
+++ b/server/router_utils.go
@@ -83,7 +83,8 @@ func readCompositeRuleID(writer http.ResponseWriter, request *http.Request) (
 		return
 	}
 
-	ruleID = ctypes.RuleID(ruleIDParam)
+	//trim the .report from module part if present
+	ruleID = ctypes.RuleID(trimDotReportFromRuleID(ruleIDParam))
 	return
 }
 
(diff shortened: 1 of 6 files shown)


server/router_utils.go
```diff
@@ -83,7 +83,8 @@ func readCompositeRuleID(writer http.ResponseWriter, request *http.Request) (
 		return
 	}
 
-	ruleID = ctypes.RuleID(ruleIDParam)
+	//trim the .report from module part if present
```
tisnik, line 86: space :)

tisnik (approved):
Seems legit

server/router_utils.go
epapbak: True, true. Adding that.
BTW the `writer` argument is not used, so I'll remove it from the function.

dpensi (approved):

PR #868, opened 2022-02-06
v2/rule takes user disabled clusters into account for impacted clusters count

Bee-lee:
# Description

Modifies /rule endpoint so that it accounts for user disabled clusters for single rules
- sends list of active clusters to aggregator to retrieve the list of disabled rules for them.
- takes disabled clusters into account when calculating the total impacted count

Also modifies config-devel, so it uses the aggregator cluster list fallback.

Fixes CCXDEV-7096

## Type of change

## Testing steps
`make before_commit` + alongside aggregator

## Checklist

### DIFF
--- a/abcgo.sh
+++ b/abcgo.sh
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-threshold=62
+threshold=63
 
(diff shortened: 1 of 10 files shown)


tisnik (approved):
It looks ok

PR #866, opened 2022-02-04
Look for rule module with '.report' in rule enable/disable tables

epapbak:
# Description

Rules disabled using v1 enable/disable endpoints include '.report' in the module, so we need this trick to make advisor's rule-related operations more consistent.

Fixes [CCXDEV-7183](https://issues.redhat.com/browse/CCXDEV-7183)

## Type of change

- Bug fix (breaking-change)

## Testing steps

- tested locally
- updated corresponding UTs

## Checklist
* [x] `make before_commit` passes
* [ ] updated documentation wherever necessary
* [x] added or modified tests if necessary
* [ ] updated schemas and validators in [insights-data-schemas](https://github.com/RedHatInsights/insights-data-schemas) in case of input/output change

PR #873, opened 2022-02-21
fix /rule and /clusters behaviour regarding single disabled rules/clusters

Bee-lee:
# Description
- recently, it was decided that OCP Advisor would use the v1 enable/disable endpoints the same way OCM does (sending g `.
report` as a suffix to the rule module name)
- this was partially taken into account by [pull/866](https://github.com/RedHatInsights/insights-results-smart-proxy/pull/866), but wasn't taken into account on the other v2 endpoints that deal with v1-related rule disables.

This explains why it was working before, because if you tested the enable/disable functionality via Advisor UI prior to this change, the UI sent the request with rule module without the `.report` suffix, therefore it was taken into account in Recommendations and Clusters lists.

This also explains why it was working on prod, but not in stage, even though the backend version was the same. It's because the UI wasn't updated on prod yet, therefore it was still sending rule module without the suffix, whereas on stage, it was already sending the suffix, effectively breaking the functionality on 2/3 endpoints.

## Type of change

## Testing steps
hard to test and debug because of special hardcoded values 1) from frontend 2) from aggregator, both of which we're unable to test in smart-proxy

## Checklist

### DIFF
--- a/server/handlers_test.go
+++ b/server/handlers_test.go
@@ -1625,7 +1625,7 @@ func TestHTTPServer_RecommendationsListEndpoint2Rules_ImpactingMissing1RuleDisab
 			"rules":[
 				{
 					"ClusterID": "%v",
-					"RuleID": "%v",
+					"RuleID": "%v.report",
 					"ErrorKey": "%v"
 				}
 			],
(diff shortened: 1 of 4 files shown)


tisnik (approved):
It looks ok!

juandspy (approved):

PR #884, opened 2022-04-11
fix last_checked_at behaviour in v2/cluster/{clusterId}/reports and v2/clusters endpoints

Bee-lee:
# Description
- `v2/cluster/{clusterId}/reports` is no longer ommitting `last_checked_at` when it shouldn't
- `v2/clusters` is now ommitting `last_checked_at` when it should

Fixes CCXDEV-7218, CCXDEV-7761

## Type of change

## Testing steps
tested behaviour with archives with and without rule hits

## Checklist

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,3 +1,17 @@
+# Copyright 2022 Red Hat, Inc
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


tisnik (commented):
It will affects all clients, right? By omitting the (sometimes empty) timestamp attribute?

Bee-lee:
@tisnik yes, Advisor is expecting this change and the new client (IO) is also expecting and waiting on this fix

tisnik (approved):
LGTM

PR #888, opened 2022-04-27
Data pipelines unification

tisnik:
# Description

Data pipelines unification

## Type of change

- Documentation update

## Testing steps

N/A

## Checklist


Bee-lee (approved):

joselsegura (approved):
LGTM

Maybe it can be simplified a bit (using same ccx-data-pipeline archives handler, adding internal rules and using 2 different consumer groups for each aggregator).

Probably it will be worst for the aggregator POV to distinguish between internal and external...

tisnik:
> 

thank you @joselsegura . It basically needs some "hard" changes in the code + probably need re-review from sec. team, which might be problematic.

PR #902, opened 2022-06-08
fix edge case in clusters_detail when all clusters are managed and requested rule is not

Bee-lee:
# Description
Fix a bug where smart-proxy used the aggregator fallback solution to retrieve a list of clusters when the list of active clusters was empty, but valid (clusters were filtered out).

Fixes 8109

## Type of change

## Testing steps
`make before_commit` + checked behaviour with hacked code

## Checklist


tisnik (approved):
Seems ok'

juandspy (approved):

PR #904, opened 2022-06-21
return "impacted" field

dpensi:
# Description

return "impacted" field

Fixes # CCXDEV-7860

## Type of change

## Testing steps

-

## Checklist


tisnik (commented):
I'd increase ReadTimeout to 1 minute or so

juandspy (approved):

PR #910, opened 2022-07-18
return impacted field

dpensi:
# Description

new field `impacted` on endpoint `cluster/{cluster}/reports`

Fixes CCXDEV-7859

## Type of change

## Testing steps

N/A

## Checklist

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -894,11 +894,9 @@ func (server HTTPServer) buildReportEndpointResponse(
 	)
 	log.Info().Msgf("Cluster ID: %v; visible rules %d, no content rules %d, disabled rules %d", clusterID, len(visibleRules), noContentRulesCnt, disabledRulesCnt)
 
-	if err != nil {
-		if _, ok := err.(*content.RuleContentDirectoryTimeoutError); ok {
-			handleServerError(writer, err)
-			return nil, 0, err
-		}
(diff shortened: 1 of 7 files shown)


server/server.go
```diff
@@ -894,11 +894,9 @@ func (server HTTPServer) buildReportEndpointResponse(
 	)
 	log.Info().Msgf("Cluster ID: %v; visible rules %d, no content rules %d, disabled rules %d", clusterID, len(visibleRules), noContentRulesCnt, disabledRulesCnt)
 
-	if err != nil {
-		if _, ok := err.(*content.RuleContentDirectoryTimeoutError); ok {
-			handleServerError(writer, err)
-			return nil, 0, err
-		}
+	if _, ok := err.(*content.RuleContentDirectoryTimeoutError); ok {
```
tisnik, line 897: does not it throw `nil pointer`?

tisnik (approved):
LGTM, just two comments

server/server.go
dpensi: no, `ok` will be false in case of nil https://play.golang.com/p/OOj8m7WJobe , in fact it was suggested by my linter to remove the nil check https://staticcheck.io/docs/checks#S1020

tisnik: cool, TY

juandspy (approved):

PR #912, opened 2022-07-25
add auth functionality for new token structure (org_id top level), check empty account_number

Bee-lee:
# Description

I'm adding this without unit tests for now because I realized we don't have the xrh auth type covered almost at all... So it's a lot of tests. Let's see how it behaves on stage, since the testing started today.

Fixes # (issue)

## Type of change

## Testing steps
local, auth token retrieved from stage

## Checklist


juandspy (approved):

tisnik (approved):
LGTM

PR #919, opened 2022-09-07
allow anemic tenants (users without `account_number`) through the API

Bee-lee:
# Description
This is the final PR for the EBS/org_id migratoin. This enables anemic tenants (users without `account_number`) through the API.
All usages of `account_number` in smart-proxy and aggregator are gone, so we can safely let them through.

I'm repeating again that we haven't had a single user like that in over a month on stage and several days in prod.

Fixes # (issue)

## Type of change

## Testing steps
with locally modified tokens according to the new spec and with tokens copied directly from crc.

## Checklist


server/auth.go
```diff
@@ -124,11 +124,16 @@ func (server *HTTPServer) Authentication(next http.Handler, noAuthURLs []string)
 			log.Error().Msg("org_id not found on top level in token structure (old format)")
 		}
 
-		if tk.Identity.AccountNumber == "" || tk.Identity.Internal.OrgID == 0 {
-			msg := fmt.Sprintf("error retrieving requester data from token. org_id [%v], account_number [%v], user data [%+v]",
-				tk.Identity.Internal.OrgID,
+		if tk.Identity.AccountNumber == "" || tk.Identity.AccountNumber == "0" {
```
tisnik, line 127: oh so "0" is special?

tisnik (approved):
looks correct as far as I can say

server/auth.go
Bee-lee: they mentioned in can be missing altogether from the token, but I don't trust anyone so I added the "0" condition "just in case"..

PR #927, opened 2022-09-22
add fallback variable

dpensi:
# Description
add environment variable to allow/disallow fall back to aggregator when ams client is not reachable

Fixes CCXDEV-9256

## Type of change

## Testing steps

na

## Checklist


tisnik (approved):
LGTM

PR #935, opened 2022-10-10
extend v1/report functionality to differentiate between requests coming from insights-operator

Bee-lee:
# Description
We were asked to modify the v1/report endpoint to differentiate between requests coming from IO and all other requesters. Any other consumer of this endpoint MUSTN'T be affected by this change. 
 
The reason being is that IO shipped with OCP v4.10 and older still uses this v1 endpoint, whereas in 4.11, IO was updated to use the new v2 endpoint, which already takes managed clusters and rules into account. This creates an inconsistency in the data the customers can see in different places (Web console didn't match Advisor). 

Because this only affects managed clusters and customers cannot simply upgrade their clusters to 4.11, we need to accommodate for this case ourselves.

In theory, we could remove this after all customers are on 4.11 and newer, but we would never know if someone may have installed an older version for some reason, so this ugly hack is probably here to stay....

Fixes https://issues.redhat.com/browse/CCXDEV-9393
https://issues.redhat.com/browse/CCXDEV-9386

## Type of change

## Testing steps

Please describe how the change was tested locally. If, for some reason, the testing was not done or not done fully, please describe what are the testing steps.

## Checklist

### DIFF
--- a/server/handlers_test.go
+++ b/server/handlers_test.go
@@ -636,6 +636,114 @@ func TestHTTPServer_ReportEndpoint_WithOnlyOSDEndpoint(t *testing.T) {
 	}, testTimeout)
 }
 
+func TestHTTPServer_ReportEndpoint_InsightsOperatorUserAgentManagedCluster(t *testing.T) {
+	defer content.ResetContent()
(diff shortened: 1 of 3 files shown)


joselsegura (approved):
LGTM

PR #1005, opened 2023-02-09
update docker image

dpensi:
# Description

update docker image to use golang 1.18

Fixes CCXDEV-10112

## Type of change

- Security fix in dependent library (no changes in the code)
- 
## Testing steps

- n/a

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM registry.redhat.io/rhel8/go-toolset:1.17 AS builder
+FROM registry.redhat.io/rhel8/go-toolset:1.18.9 AS builder
 
 COPY . .
 


Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM registry.redhat.io/rhel8/go-toolset:1.17 AS builder
+FROM registry.redhat.io/rhel8/go-toolset:latest AS builder
```
joselsegura, line 15: I think that is a good idea to stick to an specific go version. If not, in the future we can realize that, if our code is not compatible with 1.19, some day it starts failing without any reason in our end

joselsegura (changes requested):

tisnik (approved):
ok

joselsegura (approved):

PR #1021, opened 2023-04-05
Expose AMS client token so it can be overwritten by tests in ephemeral

ikerreyes:
# Description

Testing in ephemeral with a mock instance of AMS requires a valid auth mechanism. Thus, it requires a valid client and secret or a token. Token is used by tests, and given that the goal is to enable ephemeral testing, we follow the same approach.

It is set as an env. var. instead of a secret to make it easier to set it via `bonfire` command flags and thus not requiring deploying a new secret and restarting the service.

## Type of change

- Configuration update

## Testing steps

None

## Checklist


tisnik (approved):
ok

epapbak (approved):
works for me

PR #1052, opened 2023-05-31
[CCXDEV-11073] OpenAPI specification for on-demand data gathering endpoints

epapbak:
# Description

- Add specification for the new endpoints in v2/openapi.json

Fixes CCXDEV-11073

## Type of change

## Testing steps

CI

## Checklist

### DIFF
--- a/server/api/v2/openapi.json
+++ b/server/api/v2/openapi.json
@@ -89,8 +89,7 @@
                       }
                     },
                     "status": {
-                      "type": "string",
-                      "example": "ok"
+                      "$ref": "#/components/schemas/statusResponse"
                     }
                   }
                 }
@@ -127,9 +126,8 @@
           {
             "example": "34c3ecc5-624a-49a5-bab8-4fdc5e51a266",
             "name": "clusterId",
-            "description": "ID of the cluster which must conform to UUID format.",
             "schema": {
-              "type": "string"
+              "$ref": "#/components/schemas/clusterId"
             },
             "in": "path",
             "required": true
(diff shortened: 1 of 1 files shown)


server/api/v2/openapi.json
```diff
@@ -1744,6 +1988,32 @@
           }
         }
       },
+      "simplifiedReport": {
+        "type": "array",
+        "items": {
+          "type": "object",
+          "properties": {
+            "rule_fqdn": {
```
Bee-lee, line 1996: is rule_fqdn an agreed name? IIRC It's not used anywhere in our API and people always have different understanding of this name. Most people expect the error key too if it's called "fully qualified......"

server/api/v2/openapi.json
```diff
@@ -357,6 +346,254 @@
...
+            "description": "Invalid request (e.g cluster ID with unexpected format)"
+          },
+          "404": {
+            "description": "Cluster not found"
+          }
+        }
+      },
+      "post": {
+        "summary": "Filtered list of stored requests for given cluster ID",
+        "description": "A list of requests for given cluster, filtered by the provided request IDs, if any.",
```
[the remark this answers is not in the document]
epapbak: With the specification as is, it should return a `400 Bad request` if the POST endpoint is used without providing at least one request ID.

server/api/v2/openapi.json
epapbak: Let me run that through the right people and come back to you

epapbak: We have some inconsistencies when it comes to naming rules (with or without error key, with '.report', etc) due supporting different clients, but we kind of agreed on using `rule_fqdn|erorr_key` combination and call it `rule_selector` (e.g v2 `clusters_detail` endpoint).

The `rule_fqdn` name comes from the [rules specification](https://docs.google.com/document/d/1bGwR1J7bezb2k-KbJ5lVtEd4ftq_9WSVK6vd_3MllwY/edit#heading=h.i00d53mlijgv) itself.

Bee-lee (approved):

tisnik (approved):
ok

PR #1054, opened 2023-06-02
implement `clusters/{cluster}/request/{request_id}/status` endpoint

Bee-lee:
# Description
- uses moved Redis client from utils repo
- implements `clusters/{cluster}/request/{request_id}/status` endpoint
- function utilizing Redis `GetRequestIDsForClusterID` will be used in other endpoints as well

## Type of change

- REST API tests

## Testing steps
local testing with real Redis

NOTE: `make before_commit` doesn't pass locally for some reason, errcheck returns ambiguous error `error: failed to check packages: errors while loading package github.com/RedHatInsights/insights-results-smart-proxy_test` but no new package was added ... let's see what CI has to say

## Checklist

### DIFF
--- a/server/endpoints_v2.go
+++ b/server/endpoints_v2.go
@@ -55,6 +55,13 @@ const (
 	// ContentV2 returns all the static content available for the user
 	ContentV2 = "content"
 
+	// Endpoints to manipulate with simplified rule results stored
+	// independently under "tracker_id" identifier in Redis
+
+	// StatusOfRequestID should return status of processing one given
+	// request ID
+	StatusOfRequestID = "clusters/{cluster}/request/{request_id}/status"
+
 	// Endpoints to acknowledge rule and to manipulate with
 	// acknowledgements.
 
@@ -104,6 +111,9 @@ func (server *HTTPServer) addV2EndpointsToRouter(router *mux.Router) {
 	// Rules related endpoints
 	server.addV2RuleEndpointsToRouter(router, apiV2Prefix, aggregatorBaseEndpoint)
 
+	// Endpoints requiring Redis to work
+	server.addV2RedisEndpointsToRouter(router, apiV2Prefix)
+
 	// Prometheus metrics
 	router.Handle(apiV2Prefix+MetricsEndpoint, promhttp.Handler()).Methods(http.MethodGet)
 
@@ -115,6 +125,12 @@ func (server *HTTPServer) addV2EndpointsToRouter(router *mux.Router) {
 		openAPIv2URL,
 		httputils.CreateOpenAPIHandler(server.Config.APIv2SpecFile, server.Config.Debug, true),
 	).Methods(http.MethodGet)
+}
+
+// addV2RedisEndpointsToRouter method registers handlers for endpoints that depend on our Redis storage
+// to provide responses.
+func (server *HTTPServer) addV2RedisEndpointsToRouter(router *mux.Router, apiPrefix string) {
+	router.HandleFunc(apiPrefix+StatusOfRequestID, server.readStatusOfRequestID).Methods(http.MethodGet)
 }
 
 // addV2ReportsEndpointsToRouter method registers handlers for endpoints that
--- a/server/endpoints_v2.go
+++ b/server/endpoints_v2.go
@@ -130,7 +130,7 @@ func (server *HTTPServer) addV2EndpointsToRouter(router *mux.Router) {
 // addV2RedisEndpointsToRouter method registers handlers for endpoints that depend on our Redis storage
 // to provide responses.
 func (server *HTTPServer) addV2RedisEndpointsToRouter(router *mux.Router, apiPrefix string) {
-	router.HandleFunc(apiPrefix+StatusOfRequestID, server.readStatusOfRequestID).Methods(http.MethodGet)
+	router.HandleFunc(apiPrefix+StatusOfRequestID, server.getRequestStatusForCluster).Methods(http.MethodGet)
 }
 
 // addV2ReportsEndpointsToRouter method registers handlers for endpoints that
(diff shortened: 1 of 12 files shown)


server/endpoints_v2.go
```diff
@@ -117,6 +127,12 @@ func (server *HTTPServer) addV2EndpointsToRouter(router *mux.Router) {
 	).Methods(http.MethodGet)
 }
 
+// addV2RedisEndpointsToRouter method registers handlers for endpoints that depend on our Redis storage
+// to provide responses.
+func (server *HTTPServer) addV2RedisEndpointsToRouter(router *mux.Router, apiPrefix string) {
+	router.HandleFunc(apiPrefix+StatusOfRequestID, server.readStatusOfRequestID).Methods(http.MethodGet)
```
epapbak, line 133: I called the method "getRequestStatusForCluster" in the openAPI. I don't know if we are using any docs generator or something similar, in which case it could be confusing. just FYI

server/handlers_v2.go
```diff
@@ -51,6 +51,13 @@ const (
 	// OkMsg is in status field with HTTP 200 response
 	OkMsg       = "ok"
 	selectorStr = "selector"
+	// StatusProcessed is a message returned for already processed reports stored in Redis
+	StatusProcessed = "processed"
+	// RequestsForClusterNotFound is a message returned when no request IDs were found for a given clusterID
+	RequestsForClusterNotFound = "Requests for cluster not found"
```
epapbak, line 57: "No requests found for given cluster" sounds more accurate than ""Requests for cluster not found"". WDYT?

services/redis.go
```diff
@@ -18,79 +18,85 @@ package services
...
+		var keys []string
+		var err error
+		keys, cursor, err = redis.Client.Connection.Scan(ctx, cursor, scanKey, 0).Result()
+		if err != nil {
+			log.Error().Err(err).Msgf("failed to execute SCAN command for key '%v' and cursor '%d'", scanKey, cursor)
+			return nil, err
+		}
+
+		// get last part of key == request_id
+		for _, key := range keys {
```
epapbak, line 89: probably more efficient to initialise the slice and insert with indexing.

                requestIDs = make(..., len(keys), len(keys))
		for idx, key := range keys {
			keySliced := strings.Split(key, ":")
			requestID := keySliced[len(keySliced)-1]
			requestIDs[idx] = types.RequestID(requestID)
		}

epapbak (commented):
LGTM, but there are some changes I'd do

server/handlers_test.go
```diff
@@ -3832,7 +3832,7 @@ func TestHTTPServer_GroupsEndpoint(t *testing.T) {
 			StatusCode: http.StatusOK,
 			Body:       expectedBody,
 		})
-	}, testTimeout)
+	}, 30*time.Second)
```
juandspy, line 3835: why do you use `30s` instead of the `testTimeout`?

server/handlers_v2.go
juandspy: +1

types/types.go
```diff
@@ -48,6 +48,10 @@ type OrgID = types.OrgID
 // ImpactingFlag controls the behaviour of 'impacting' param on GET /rule/
 type ImpactingFlag int
 
+// RequestID is used to store the request ID supplied in input Kafka records as
+// a unique identifier of payloads. Empty string represents a missing request ID.
+type RequestID string
```
tisnik, line 53: probably later it should go to `types` as it is used in Aggregator too (and in IO operator as well).

tisnik (approved):
LGTM (but please look at other comments, they are relevant)

juandspy (approved):

server/handlers_test.go
Bee-lee: because these two test cases are very unstable and sometimes it takes more than 10s. I found that 30s was safe enough for this while not being too long in case it fails for a real reason.

server/handlers_v2.go
Bee-lee: @tisnik is someone already using the mock API to expect this? I was following these messages from the mock so that it's not different in case someone already expects it.

services/redis.go
Bee-lee: @epapbak we don't know the total length though, SCAN doesn't return the whole dataset, it returns a "page" with a default length of 10 and it returns a cursor which needs to be used in subsequent SCAN executions to get the whole dataset, hence the while loop.

server/endpoints_v2.go
Bee-lee: @epapbak if you mean the `operationId` in the OpenAPI spec, it doesn't correspond to the names of our functions. I just checked and out of 42 (nice) `operationId`, only 1 actually matches the handler function name (GET /rule). AFAIK it's only used in some way by IQE tests, but I might be wrong

btw you're missing an operationId definition [here ](https://github.com/RedHatInsights/insights-results-smart-proxy/blob/master/server/api/v2/openapi.json#L400), I can add it in the next PR if you want

server/handlers_v2.go
tisnik: @Bee-lee not now, as IO team work on on-demand data gathering is a bit delayed. We can use the proposed message, not a problem at all.

services/redis.go
tisnik: + it does not run in transaction, so it is not possible to tell the exact number of records (AFAIK - dunno if/how Redis will work)

server/endpoints_v2.go
Bee-lee: updated for the peace of mind :)

server/handlers_v2.go
Bee-lee: ok thanks! done

server/endpoints_v2.go
Bee-lee: FYI I'm adding the missing `operationId` in the next PR

PR #1056, opened 2023-06-05
[CCXDEV-10263] Simplify rule content cache update mechanism

epapbak:
# Description

With this change, we improve the cache in the following way:
- The cache is updated once all the retrieved rules are parsed and added to a temporary RuleWithContentStorage. Every time the content is retrieved, the current cache is swapped with the temporary storage once it is fully prepared. Meanwhile, the previous cache is used to respond to requests related with rules content.
- The `content.ResetContent` function is no longer needed, as we are replacing the `RuleWithContentStorage` by assigning it to a new cache and discarding the previously existing one.
- The  `content.ResetContent` function has been moved to the `export_test.go` file of the content package as a helper for unit tests, and is no longer used in other packages.
- There is no longer a need for using mutexes as there is no simultaneous read and write, which should make the cache even more efficient.

Fixes CCXDEV-10263

## Type of change

## Testing steps

- CI
- Tested the failing scenario against the service running locally with the fix

## Checklist

### DIFF
--- a/content/content.go
+++ b/content/content.go
@@ -35,13 +35,9 @@ var (
 	ruleContentDirectory      *ctypes.RuleContentDirectory
 	ruleContentDirectoryReady = sync.NewCond(&sync.Mutex{})
 	stopUpdateContentLoop     = make(chan struct{})
-	rulesWithContentStorage   = RulesWithContentStorage{
-		rules:                      map[ctypes.RuleID]*ctypes.RuleContent{},
-		rulesWithContent:           map[ruleIDAndErrorKey]*types.RuleWithContent{},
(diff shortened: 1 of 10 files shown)


epapbak:
> ci.int.devshift.net PR build failed to push results to ibutsu, but all tests passed.

Same thing. 

> 11:43:35 PASSED                                                                   [100%]Saved results archive to 83d393db-dfaf-48b9-a281-cc249c93be00.tar.gz

tisnik (approved):
it looks much cleaner now

juandspy (approved):

Bee-lee (approved):
LGTM

PR #1059, opened 2023-06-14
`x-rh-identity` tests 1st part

Bee-lee:
# Description
1st part of adding UT coverage for `x-rh-identity` auth header type.

JWT auth type is not used in any environment anymore, so until now, we practically had zero coverage for the auth type we're actually using.

Fixes # (issue)

## Type of change

## Testing steps
`make before_commit`

## Checklist

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -32,6 +32,10 @@ import (
 )
 
 const (
+	// JWTAuthTokenHeader reprsents the name of the header used in JWT authorization type
+	JWTAuthTokenHeader = "Authorization"
+	// XRHAuthTokenHeader reprsents the name of the header used in XRH authorization type
+	XRHAuthTokenHeader = "x-rh-identity"
 	// #nosec G101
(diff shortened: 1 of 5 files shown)


joselsegura (approved):

server/auth.go
```diff
@@ -217,7 +223,7 @@ func (server *HTTPServer) getAuthTokenHeader(w http.ResponseWriter, r *http.Requ
 	} else {
```
juandspy, line 223: shall we convert this `else` into an `else if` and use the `else` to return `false` as the authentication type is unknown?

juandspy (approved):

server/auth.go
Bee-lee: good point, but we have a task in the backlog that I'm planning to do to completely drop the `jwt` part because it's really not used anywhere at all

juandspy: ok!

tisnik (approved):
nice one!

PR #1064, opened 2023-06-26
[CCXDEV-10956] Get the timestamp from data-eng instead of generating a new one each time

juandspy:
# Description

Currently we are generating a timestamp of `now()` in every request. As we are implementing a cache in data-eng, let's use the timestamp that data-eng will return.

## Type of change

## Testing steps

Locally.

## Checklist

### DIFF
--- a/types/upgrade_risks_predictions.go
+++ b/types/upgrade_risks_predictions.go
@@ -36,6 +36,13 @@ type UpgradeRisksPredictors struct {
 	OperatorConditions []OperatorCondition `json:"operator_conditions"`
 }
 
+// DataEngResponse is the response received from the data-eng service
+type DataEngResponse struct {
+	Recommended     bool                   `json:"upgrade_recommended"`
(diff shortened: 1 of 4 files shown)


Bee-lee (approved):
looks ok

types/upgrade_risks_predictions.go
```diff
@@ -36,6 +36,13 @@ type UpgradeRisksPredictors struct {
 	OperatorConditions []OperatorCondition `json:"operator_conditions"`
 }
 
+// DataEngResponse is the response received from the data-eng service
```
tisnik, line 39: this should probably go into `types` repository

tisnik (approved):
LGTM - just comment about new structure

types/upgrade_risks_predictions.go
juandspy: it's used just in this repo. Is it needed to be moved there?

PR #1081, opened 2023-07-10
fix incorrect behaviour of `GetRequestIDsForClusterID`

Bee-lee:
# Description
- fixes a bug where `GetRequestIDsForClusterID` was also matching keys which are used for simplified reports, because they share the same prefix. Initially I thought it cannot be done via the MATCH pattern, because regular glob doesn't support exclusion patterns, however Redis uses "glob-style patterns", which allow exclusion functionality (see https://redis.io/docs/manual/keyspace/) 
  - this unfortunately cannot be covered by UTs because we use redis-mock, therefore we tell the mock what to return
- extends logs for further debugging of another Redis issue

Fixes CCXDEV-11338

## Type of change

## Testing steps
tested with local Redis, will be tested on stage properly

## Checklist

### DIFF
--- a/services/redis.go
+++ b/services/redis.go
@@ -104,7 +104,7 @@ func (redis *RedisClient) GetRequestIDsForClusterID(
 	ctx := context.Background()
 
 	scanKey := fmt.Sprintf(RequestIDsScanPattern, orgID, clusterID)
-	log.Debug().Str("Scan key", scanKey).Msg("Key to retrieve item from Redis")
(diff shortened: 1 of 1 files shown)


tisnik (approved):
LGTM

joselsegura (approved):

epapbak (approved):
Add to the non-existent (yet) list of BDD tests for this service =)

juandspy (approved):

PR #1082, opened 2023-07-11
Redis fixes

Bee-lee:
# Description
- increasing the SCAN command batch size (`count` parameter) to 1000 from the default 10 because the SCAN command currently takes ~10s in production

pic.related: benchmark on 5M keys with SCAN COUNT 
image

Fixes # (issue)

## Type of change

## Testing steps
with local Redis

## Checklist


tisnik (approved):
LGTM (probably later we might add it as a config option)

PR #1088, opened 2023-07-21
[CCXDEV-10215] CI: Use openapi-generator-cli docker image instead of installing it

epapbak:
# Description

Use the docker image provided by the OpenApiTools organization instead of installing the `openapi-generator-cli` every time a PR is verified.

The installation seems to fail quite often due to downtimes of the repository used to verify and download the jar file used by the verification tool.

The issue does not seem to have been fixed yet. See https://github.com/OpenAPITools/openapi-generator-cli/issues/680 and https://github.com/OpenAPITools/openapi-generator/issues/9349. So using the docker image allows us to use the tool without interacting with the "failing" repository.

Fixes CCXDEV-10215

## Type of change

## Testing steps

CI

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,15 +33,14 @@ jobs:
         - env
         - bash <(curl -s https://codecov.io/bash)
     - stage: openapi-checks
-      language: node_js
-      node_js:
-        - 17
-        - node
+      services:
+        - docker
+      before_install:
+        - docker pull openapitools/openapi-generator-cli
       script:
-        - npm install -g @openapitools/openapi-generator-cli
-        - openapi-generator-cli validate -i server/api/v1/openapi.json
-        - openapi-generator-cli validate -i server/api/v2/openapi.json
-        - openapi-generator-cli validate -i server/api/dbg/openapi.json
+        - docker run --rm -v ${PWD}:/local openapitools/openapi-generator-cli validate -i /local/server/api/v1/openapi.json
+        - docker run --rm -v ${PWD}:/local openapitools/openapi-generator-cli validate -i /local/server/api/v2/openapi.json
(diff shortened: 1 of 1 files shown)


Bee-lee (approved):

tisnik (approved):
Let's try it

.travis.yml
```diff
@@ -33,15 +33,14 @@ jobs:
...
         - bash <(curl -s https://codecov.io/bash)
     - stage: openapi-checks
-      language: node_js
-      node_js:
-        - 17
-        - node
+      services:
+        - docker
+      before_install:
+        - docker pull openapitools/openapi-generator-cli
```
ikerreyes, line 39: Isn´t there a risk that docker blocks the downloads? They have a very small limit for anonymous accounts

epapbak: There is. I thought we should be fine because of the following: If you check out the [insights tab of our Travis](https://app.travis-ci.com/github/RedHatInsights?tab=insights), we see an average total of **600 jobs per week** for all the repositories, while the rate limit is **100 pulls every 6 hours**.

But I admit I couldn't figure out what kind of pool is used for the runners, so I'm looking into ways to not be bitten by the rate limiting.

So yes, there is a risk hat docker blocks the downloads, but IMHO it won't affect us as much as the current issue with the maven repository.

weekly insights: image

monthly insights: 
image

ikerreyes: Thanks for the insights. I would still be worried, because I guess Travis instances share IP addresses, so it is not how much you download but how much everyone does. But if you think the benefits overcome the drawbacks, let's do it.

ikerreyes (approved):

PR #1098, opened 2023-08-04
fix nonsensical logging event (never worked)

Bee-lee:
# Description
- fixes `"error":"json: cannot unmarshal object into Go value of type []types.RuleOnReport"`
  - this could've never worked before (almost 2y), first of all it's not a list, second of all it's a totally different struct...
  - the error message was displayed for every single request for relevant endpoints (about 2-3 error messages **PER SECOND**)

## Type of change

## Testing steps
locally with relevant endpoints + aggregator response

## Checklist

### DIFF
--- a/server/utils.go
+++ b/server/utils.go
@@ -24,9 +24,9 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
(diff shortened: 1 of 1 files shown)


tisnik (approved):
okay

ikerreyes (approved):

JiriPapousek (approved):

PR #1104, opened 2023-09-06
fix possible duplicate records returned from AMS API. cluster UUID is unique for us.

Bee-lee:
# Description
AMS API can sometimes return duplicate records even though `page_size=-1` (unlimited page size). Possibly due to some internal paging mechanism inside AMS API. Because we didn't modify the cluster list returned from AMS, these duplicated clusters have been showing in the Insights Advisor for OpenShift application (noticed by tdosek in his org). 

this fix included:
- unlinking my c.r.c account from my organization
- linking my account to known problematic organization (tdosek's org)
- testing behaviour on prod to check if it's backend/frontend issue
- retrieving access token for said organization
- modifying `amsclient` auth locally to accept access tokens instead of client ID and secret
- fix in this PR
- re-test with modified amsclient
- re-link my proper org to my account

Fixes https://issues.redhat.com/browse/CCXDEV-11659

## Type of change

## Testing steps
mentioned in description

## Checklist

### DIFF
--- a/amsclient/amsclient.go
+++ b/amsclient/amsclient.go
@@ -242,6 +242,8 @@ func (c *amsClientImpl) executeSubscriptionListRequest(
 	clusterInfoList []types.ClusterInfo,
 	err error,
 ) {
+	uniqueClusterMap := make(map[string]bool)
+
 	for pageNum := 1; ; pageNum++ {
 		var err error
 		subscriptionListRequest = subscriptionListRequest.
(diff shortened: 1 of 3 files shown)


amsclient/amsclient.go
```diff
@@ -242,6 +242,8 @@ func (c *amsClientImpl) executeSubscriptionListRequest(
 	clusterInfoList []types.ClusterInfo,
 	err error,
 ) {
+	uniqueClusterMap := make(map[string]bool)
```
tisnik, line 245: you can use a map of empty structs which is optimized (or should be according to Go spec/idioms).

tisnik (commented):
Looks ok, but I'd use map of empty structs as it is the idiom for sets in Go

amsclient/amsclient.go
Bee-lee: done

tisnik (approved):
LGTM

juandspy (approved):

PR #1105, opened 2023-09-06
add rule acknowledge functionality to `v2/clusters_detail`

Bee-lee:
# Description
`v2/clusters_detail` wasn't using information about rule acks. This isn't a problem for Advisor because the call to `/clusters_detail` was controlled by the response from `/rule/{rule_id}` which was called prior to this:
- if `/rule/{rule_id}` returned that the rule is acknowledged, the request to `/clusters_detail` was never made == no problem
- if `/rule/{rule_id}` returned that the rule isn't acknowledged, the request to `/clusters_detail` was made and correct information was provided (from Advisor perspective).

Recently there have been some pushes to inform customers that they can use this API to make their own results/aggregations, so we need to make sure the API is consistent. This PR introduces correct handling of rule acknowledges. Disabling single clusters takes precedence over general rule acknowledges.

Fixes # (issue)

## Type of change

- REST API tests

## Testing steps
new behaviour locally

## Checklist


juandspy (approved):

tisnik (approved):
LGTM. Probably the selector split is implemented somewhere else already (dunno, in utils? I'm too lazy to look ;)

juandspy:
@Bee-lee this is going to fail because of [this](https://redhat-internal.slack.com/archives/C01A4DTTZN1/p1694500288185649)

PR #1111, opened 2023-09-12
`v1/organizations/{organization}/clusters` now uses AMS API functionality to get cluster list

Bee-lee:
# Description
`v1/organizations/{organization}/clusters` was practically unusable for very large orgs (>20k clusters) because it was retrieving clusters by string (date) comparison. Combined with the large number of rows in the table, it took a very long time to retrieve the results and triggered a gateway timeout.

- endpoint is not utilized by any real consumer, only tests (visible in metrics)
  - removing the endpoint is still not a good idea
- already mentioned in OpenAPI spec that v2/clusters should be used instead
- historically this endpoint was used before we introduced the AMS API
- aggregator fallback functionality is kept

Fixes CCXDEV-11657

## Type of change

- REST API tests

## Testing steps
`make before_commit`

## Checklist


tisnik (approved):
LGTM

JiriPapousek (approved):

PR #1127, opened 2023-09-22
Renamed to use the right identifier

tisnik:
# Description

Renamed to use the right identifier

## Type of change

## Testing steps

N/A

## Checklist


matysek (commented):
/lgtm

PR #1217, opened 2024-02-02
Fix missing status in data engineering response

joselsegura:
# Description

Data Engineering service doesn't include an "status" key in its JSON responses, so the value will be always missing. This PR will take that into account and avoid returning empty list of predictions on success.

Fixes #CCXDEV-11669

## Type of change

## Testing steps

Tested locally

## Checklist


JiriPapousek (approved):

Bee-lee (approved):
would be nice to have API response unit tests for this functionality

PR #1223, opened 2024-02-19
[CCXDEV-12001] Update base images

juandspy:
# Description

We decided to use non pinged versions so that dependencies are updated kind of automatically with each PR we merge. This would prevent CVEs to appear if we don't update the base images quite often.

## Type of change

- Configuration update

## Testing steps

CI.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM registry.redhat.io/rhel8/go-toolset:1.18.9-8.1675807488 AS builder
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM registry.redhat.io/rhel8/go-toolset:1.18.9-8.1675807488 AS builder
+FROM registry.redhat.io/rhel8/go-toolset:1.18.9  AS builder
```
JiriPapousek, line 15: Two spaces there (instead of one)

JiriPapousek (approved):

PR #1263, opened 2024-05-03
[CCXDEV-12791] Rename insights-results-smart-proxy to ccx-smart-proxy

juandspy:
# Description

This rename is needed in order to use the smart-proxy endpoint internally from other AppSRE namespaces. We used to have different app-sre component ID and clowdapp name which made it impossible.

## Type of change

- Configuration update

## Testing steps

CI.

## Checklist


ikerreyes (approved):
don't we need to change anything in the clowdapp file?

matysek (approved):

PR #1270, opened 2024-05-27
fix missing rule content causing 500 in DVO namespace list endpoint

Bee-lee:
# Description
- missing rule content is causing a 500, instead it should simply omit the problematic rule IDs from the calculation and return the rest of the data.

Fixes [CCXDEV-12893](https://issues.redhat.com/browse/CCXDEV-12893)

## Type of change

## Testing steps
locally

## Checklist


epapbak (approved):
makes sense

juandspy (approved):

PR #1310, opened 2024-08-16
[CCXDEV-14098] Reduce error level messages using Msgf

juandspy:
# Description

Reduce error level messages using Msgf to avoid Glitchtip duplicates.

## Type of change

## Checklist


server/acks_handlers.go
```diff
@@ -401,7 +401,9 @@ func generateRuleAckMap(acks []types.SystemWideRuleDisable) (ruleAcksMap map[typ
 		if err == nil {
 			ruleAcksMap[compositeRuleID] = true
 		} else {
-			log.Error().Err(err).Msgf(compositeRuleIDError, ack.RuleID, ack.ErrorKey)
+			log.Error().Err(err).Interface(ruleIDStr, ack.RuleID).
```
epapbak, line 404: does Str(ruleIDStr, string(ruleID)) not work in these?

juandspy: yes, but I preferred to use `Interface` in order to avoid converting types. It's just a random decision, not based on nothing major...

epapbak (approved):

ikerreyes (approved):
Seems that you have refactored a number of `Msgf` but not all. I guess it is a design decision, right?

juandspy:
> Seems that you have refactored a number of `Msgf` but not all. I guess it is a design decision, right?

I've just refactored those that are `Error` level

PR #1355, opened 2024-10-21
Add RBAC client

epapbak:
# Description

- RBAC Client
- Add authorization middleware in server
- Update UTs (little bit of coupling when it comes to CreateServer method with so much args, one more refactor commit coming)
- Add `rbac` section to configuration files

Part of CCXDEV-12884

# Why this change

After some investigation around [ADR-036](https://docs.google.com/document/d/1LNun1PuvL3EFyfbSWAa6ixUAm-pJoWzQiNJj7hVjkZY/edit?tab=t.0#heading=h.nq1osedn93p8) and a quick discussion with PM, it's been decided that we should support RBAC, mainly because if a consumer of our API uses it with any kind of user, and has configured resource-based access control, we are not enforcing it as of today.

In this change, we introduce the code necessary to handle authorization, but since we don't know which consumers use what and how it can affect them, RBAC is disabled by default and we're doing some extra logging [here](https://github.com/RedHatInsights/insights-results-smart-proxy/pull/1355/files#diff-08316311bca81628634f7ed55949099773edf5af671129d2e6bcbc7dbd1c6825R119) (there's more to come in the next and last PR).

## Type of change

- Unit tests

## Testing steps

CI

## Checklist

### DIFF
--- a/auth/rbac.go
+++ b/auth/rbac.go
@@ -62,11 +62,6 @@ func (rc *rbacClientImpl) IsEnforcing() bool {
 
 // IsAuthorized checks if an account has the correct permissions to access our resources
 func (rc *rbacClientImpl) IsAuthorized(token string) bool {
-	// baseURL := "https://console.redhat.com/api/rbac/v1" // TODO: env var, separate URL from endpoint like this: https://console.redhat.com and /api/rbac/v1/access so we can include port if needed. use helper func to create url
-	// baseURL := "https://console.stage.redhat.com/api/rbac/v1"
-
-	// Here you can add additional logic to parse the response body if needed
-	// For now, we assume that a 200 response means permissions are valid
 	permissions := rc.getPermissions(token)
 	return permissions != nil
 }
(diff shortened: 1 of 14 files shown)


Jakub007d (approved):

Bee-lee (approved):
LGTM overall, ty! would be nice to have some context included, but it is in Jira after all

JiriPapousek (approved):

auth/rbac.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+		return nil
+	}
+	return nil
+}
+
+// requestAccess handles the call(s) to RBAC taking into account that the response
+// is paginated
+func (rc *rbacClientImpl) requestAccess(url, identityToken string) []types.RbacData {
+	//TODO Change return to (rbacData, err) and forward error?
```
juandspy, line 89: leftover?

config.toml
```diff
@@ -67,3 +67,8 @@ database = 0
 endpoint = "localhost:6379"
 password = ""
 timeout_seconds = 30
+
+[rbac]
+host = https://console.redhat.com/api/rbac/v1
```
juandspy, line 72: should be string

config-devel.toml
```diff
@@ -52,3 +52,8 @@ database = 0
 endpoint = "localhost:6379"
 password = ""
 timeout_seconds = 30
+
+[rbac]
+host = https://console.redhat.com/api/rbac/v1
```
juandspy, line 57: same here

auth/rbac.go
epapbak: probably =). But I have two or three  TODOs in the new code that I'll handle later on

PR #1359, opened 2024-10-23
Avoid RBAC for requests that do not require authorization

epapbak:
# Description

- Same as we do for authentication, for `HTTP OPTIONS` and metric related URLS, let's skip RBAC.
- Also skip RBAC for requests that come from ACM

Part of CCXDEV-12884

## Type of change

## Testing steps

CI

## Checklist

### DIFF
--- a/server/auth_middleware_test.go
+++ b/server/auth_middleware_test.go
@@ -325,3 +325,59 @@ func TestAuthorizationMiddleware(t *testing.T) {
 		})
 	}
 }
+
+func TestAuthorization_NoAuthURLs(t *testing.T) {
+	// Setup the server and the request
+	rbacClient := &MockRBACClient{authorized: false, enforcing: true}
+	noAuthURLs := []string{"/public-endpoint"}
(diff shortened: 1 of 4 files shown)


server/auth_middleware_test.go
```diff
@@ -317,7 +317,7 @@ func TestAuthorizationMiddleware(t *testing.T) {
 				w.WriteHeader(http.StatusOK)
 			})
 
-			authHandler := testServer.Authorization(handler)
+			authHandler := testServer.Authorization(handler, nil)
```
juandspy, line 320: can you add some test case for a no auth url?

epapbak: I have it in another commit. Lemme push it here.

epapbak: updated, with the code to skip RBAC for ACM queries too

JiriPapousek (approved):

Bee-lee (approved):

PR #1379, opened 2024-11-25
[CCXDEV-12885] Fix bug in RBAC logic + simplify

juandspy:
# Description

I found out the RBAC permissions where filtering by "openshift" whereas it should use `ocp-advisor` as stated [here](https://github.com/RedHatInsights/rbac-config/blob/master/configs/prod/permissions/ocp-advisor.json). Also, the implementation was mimicking how [cost-management](https://github.com/RedHatInsights/rbac-config/blob/master/configs/prod/permissions/cost-management.json) defines the resources, but that's not the case for `ocp-advisor`.

In addition, we were parsing the r`esourceDefinitions` as they are also part of the [RBAC API response](https://console.redhat.com/docs/api/rbac/v1#operations-Access-getPrincipalAccess). However, we do no filtering based on it, so we can skip that part.

I added a couple of unit tests to check that the parsing is right and that we have no pointer issues if the RBAC response doesn't contain enough data.

Part of #[CCXDEV-12885](https://issues.redhat.com/browse/CCXDEV-12885)

## Type of change

## Testing steps

UTs.

## Checklist


Jakub007d (approved):

auth/rbac.go
```diff
@@ -137,30 +138,31 @@ func (rc *rbacClientImpl) requestAccess(url, identityToken string) []types.RbacD
...
+		if len(splits) < 3 {
+			log.Warn().Str("ACL", acl.Permission).Msg("Unexpected RBAC response")
+			continue
+		}
+		if splits[0] != "ocp-advisor" {
+			// check the ACL is for ocp-advisor, not for other APIs
+			continue
+		}
+		resourceType := splits[1]
+		verb := splits[2]
```
Bee-lee, line 157: what's this var representing?

Bee-lee (approved):

auth/rbac.go
juandspy: the ACLs has this format: `API:RESOURCE:VERB`, for example `ocp-advisor:read-recommendations:read`, so the `verb` variable is the action the SA can make on the given resource for the given API: https://console.redhat.com/docs/api/rbac/v1#operations-Access-getPrincipalAccess

PR #1456, opened 2025-04-09
fix an edge case where an `org_id` from 3scale matches multiple internal organizations in OCM AMS API

Bee-lee:
# Description
- there is an internal `org_id` used in the OCM AMS API which has to be retrieved from the `/organizations` endpoint.
- the `amsclient` module was implemented in a way that we always expected exactly one internal organization to be returned from the endpoint
- however a single external `org_id` (which we receive in the auth token from 3scale) can represent multiple internal `org_id`s in the OCM/AMS world. This might happen during an ownership/membership transfer, ensuring the user is not left in a state with no ID and can still access their data.
- it's also possible that the external `org_id` doesn't have an internal counterpart yet. This happens when a new account is created, 3scale and SSO are already aware of it, but OCM isn't.

Both of the described edge cases should be temporary, however both of them are ultimately fixable by the user or external systems, so our API shouldn't return 5xx. Any previously problematic `org_id` that I've tried is already working correctly (i.e. only returning one org).

This PR fixes both of these special cases.

Fixes # (issue)

## Type of change

- Documentation update

## Testing steps
`make before_commit` although golangci-lint is not behaving correctly locally.

## Checklist

### DIFF
--- a/amsclient/amsclient_test.go
+++ b/amsclient/amsclient_test.go
@@ -32,19 +32,26 @@ import (
 	"github.com/RedHatInsights/insights-results-smart-proxy/types"
 )
 
+// OCM SDK encodes URLs with escaped hex characters
+// %%2C means ,
+// %%3D means =
+// %%21 means !
+// %28%27 %27%29 means (' ')
 const (
 	organizationsSearchEndpoint = "api/accounts_mgmt/v1/organizations?fields=id%%2Cexternal_id&search=external_id+%%3D+{orgID}"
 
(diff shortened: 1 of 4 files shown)


amsclient/amsclient_test.go
```diff
@@ -139,6 +146,83 @@ func TestClusterForOrganization(t *testing.T) {
...
+		StatusCode: http.StatusOK,
+		Headers: map[string]string{
+			"Content-Type": "application/json",
+		},
+		Body: helpers.ToJSONString(testdata.OrganizationResponseNoID),
+	})
+
+	_, err = c.GetClustersForOrganization(testdata.ExternalOrgID, nil, []string{})
+	assert.Error(t, err)
+	/*
```
juandspy, line 169: is this a leftover?

Bee-lee: yes, my bad, fixed now.

Jakub007d (approved):

PR #1588, opened 2025-09-04
Adding more debuging output to find out if the empty rulesID are caused by filterOutDisabledRules function

Jakub007d:
# Description

Adding more debuging output to find out if the empty rulesID are caused by filterOutDisabledRules function


ikerreyes (approved):
Should it be info or debug level?

Jakub007d:
I think it is on debug mode as i can see there bunch of debug messages already. And those logs wont make that mutch more noise as it is quite nosy already :D

